PR #451, opened 2023-12-05
Remove release downloader script

PlagueCZ:
Now that IronCore is public, there is no need for the release downloader using PAT.

Unfortunately for Dockerfile to work well (cache properly) I had to hardcode `dpservice-cli` and `prometheus-dpdk-exporter` versions. Though the upside is that the resulting Docker image will always have "approved" versions.

Since touching the Dockerfile I made a few changes to speed it up (layer reduction and better DPDK download link)

I also changed the pytest script to download the client when needed, using the version set in the Dockerfile. This is more of a convenience for developers than anything else. It should maybe also check if the already present version is not obsolete, but I left that to the developer (as that is the previous behavior anyway).


guvenc:
@PlagueCZ 
Could you please rebase this one as well ?

guvenc (approved):
LGTM. Thanks.

PR #462, opened 2023-12-20
Introduce IPv6 overlay support for non-offloading path

guvenc:
Introduced following network functions to support overlay ipv6 in non-offloading path:

-  IPv6 conntrack support.
-  IPv6 HV-HV routing.
-  IPv6 Loadbalancer support.
-  Alias v6 Prefixes and v6 LB Alias Prefixes.
-  NAT64 support.
-  TCP/UDP NAT64.
-  ICMP NAT64.
-  DHCPv6 DNS Support.
-  Scapy tests.
-  v6 Firewall.

Scapy tests can be still extended but for the beginning there is at least one scapy test to test the introduced ipv6 overlay 
network function.

Unfortunately the commits are not suitable to be reviewed separately. As there is a cleanup commit at the end of the 
series so better take a look to the complete PR diff. The changes are somehow connected anyway.

checkpatch fails because of the cpp changes.

### DIFF
--- a/hack/dp_conf.json
+++ b/hack/dp_conf.json
@@ -48,10 +48,10 @@
       "help": "set the domain name server field in DHCP responses (can be used multiple times)"
     },
     {
-		"lgopt": "dhcpv6-dns",
-		"arg": "ADDR6",
-		"help": "set the domain name server field in DHCPv6 responses (can be used multiple times)"
-	},
+      "lgopt": "dhcpv6-dns",
+      "arg": "ADDR6",
+      "help": "set the domain name server field in DHCPv6 responses (can be used multiple times)"
+    },
     {
       "lgopt": "udp-virtsvc",
       "arg": "IPv4,port,IPv6,port",
--- a/hack/dp_conf.json
+++ b/hack/dp_conf.json
@@ -48,6 +48,11 @@
       "help": "set the domain name server field in DHCP responses (can be used multiple times)"
     },
     {
+		"lgopt": "dhcpv6-dns",
+		"arg": "ADDR6",
+		"help": "set the domain name server field in DHCPv6 responses (can be used multiple times)"
+	},
+    {
       "lgopt": "udp-virtsvc",
       "arg": "IPv4,port,IPv6,port",
       "help": "map a VM-accessible IPv4 endpoint to an outside IPv6 UDP service",
--- a/include/dp_util.h
+++ b/include/dp_util.h
@@ -39,7 +39,7 @@ extern "C" {
 #define DP_TCP_PKT_FLAG_SYNACK(FLAGS) \
 	(((FLAGS) & (RTE_TCP_SYN_FLAG|RTE_TCP_ACK_FLAG)) == (RTE_TCP_SYN_FLAG|RTE_TCP_ACK_FLAG))
 
-struct dpgrpc_address {
+struct dp_ip_address {
 	uint32_t				ip_type;
 	union {
 		uint32_t			ipv4;
--- a/include/dp_util.h
+++ b/include/dp_util.h
@@ -21,6 +21,7 @@ extern "C" {
 #define DP_IFACE_PXE_MAX_LEN	32
 #define DP_LB_ID_MAX_LEN		64
 #define DP_LB_MAX_PORTS			16
+#define DP_IPV6_ADDR_SIZE		16
 
 #define DP_MAC_EQUAL(mac1, mac2) (((mac1)->addr_bytes[0] == (mac2)->addr_bytes[0]) && \
 								((mac1)->addr_bytes[1] == (mac2)->addr_bytes[1]) && \
--- a/include/dp_util.h
+++ b/include/dp_util.h
@@ -40,10 +40,10 @@ extern "C" {
 	(((FLAGS) & (RTE_TCP_SYN_FLAG|RTE_TCP_ACK_FLAG)) == (RTE_TCP_SYN_FLAG|RTE_TCP_ACK_FLAG))
 
 struct dp_ip_address {
-	uint32_t				ip_type;
+	uint16_t		ip_type;
 	union {
-		uint32_t			ipv4;
-		uint8_t				ipv6[DP_IPV6_ADDR_SIZE];
+		uint32_t	ipv4;
+		uint8_t		ipv6[DP_IPV6_ADDR_SIZE];
 	};
 };
 
--- a/include/dp_util.h
+++ b/include/dp_util.h
@@ -39,6 +39,13 @@ extern "C" {
 #define DP_TCP_PKT_FLAG_SYNACK(FLAGS) \
 	(((FLAGS) & (RTE_TCP_SYN_FLAG|RTE_TCP_ACK_FLAG)) == (RTE_TCP_SYN_FLAG|RTE_TCP_ACK_FLAG))
 
+struct dpgrpc_address {
+	uint32_t				ip_type;
+	union {
+		uint32_t			ipv4;
+		uint8_t				ipv6[DP_IPV6_ADDR_SIZE];
+	};
+};
 
 int dp_get_dev_info(uint16_t port_id, struct rte_eth_dev_info *dev_info, char ifname[IF_NAMESIZE]);
 
--- a/include/dp_vnf.h
+++ b/include/dp_vnf.h
@@ -50,9 +50,9 @@ int dp_add_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE], enum dp_vnf_type t
 const struct dp_vnf *dp_get_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE]);
 int dp_del_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE]);
 
-bool dp_vnf_lbprefix_exists(uint16_t port_id, uint32_t vni, struct dp_ip_address *pfx_ip, uint8_t prefix_len);
+bool dp_vnf_lbprefix_exists(uint16_t port_id, uint32_t vni, struct dp_ip_address *prefix_ip, uint8_t prefix_len);
 
-int dp_del_vnf_by_value(enum dp_vnf_type type, uint16_t port_id, uint32_t vni, struct dp_ip_address *pfx_ip, uint8_t prefix_len);
+int dp_del_vnf_by_value(enum dp_vnf_type type, uint16_t port_id, uint32_t vni, struct dp_ip_address *prefix_ip, uint8_t prefix_len);
 
 int dp_list_vnf_alias_prefixes(uint16_t port_id, enum dp_vnf_type type, struct dp_grpc_responder *responder);
 
--- a/include/dp_vnf.h
+++ b/include/dp_vnf.h
@@ -31,7 +31,7 @@ enum dp_vnf_type {
 } __rte_packed;  // for 'struct dp_flow' and 'struct flow_key'
 
 struct dp_vnf_prefix {
-	struct dpgrpc_address ol;
+	struct dp_ip_address ol;
 	uint8_t				  length;
 };
 
@@ -46,13 +46,13 @@ int dp_vnf_init(int socket_id);
 void dp_vnf_free(void);
 
 int dp_add_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE], enum dp_vnf_type type,
-			   uint16_t port_id, uint32_t vni, struct dpgrpc_address *pfx_ip, uint8_t prefix_len);
+			   uint16_t port_id, uint32_t vni, struct dp_ip_address *pfx_ip, uint8_t prefix_len);
 const struct dp_vnf *dp_get_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE]);
 int dp_del_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE]);
 
-bool dp_vnf_lbprefix_exists(uint16_t port_id, uint32_t vni, struct dpgrpc_address *pfx_ip, uint8_t prefix_len);
+bool dp_vnf_lbprefix_exists(uint16_t port_id, uint32_t vni, struct dp_ip_address *pfx_ip, uint8_t prefix_len);
 
-int dp_del_vnf_by_value(enum dp_vnf_type type, uint16_t port_id, uint32_t vni, struct dpgrpc_address *pfx_ip, uint8_t prefix_len);
+int dp_del_vnf_by_value(enum dp_vnf_type type, uint16_t port_id, uint32_t vni, struct dp_ip_address *pfx_ip, uint8_t prefix_len);
 
 int dp_list_vnf_alias_prefixes(uint16_t port_id, enum dp_vnf_type type, struct dp_grpc_responder *responder);
 
--- a/include/dp_vnf.h
+++ b/include/dp_vnf.h
@@ -7,6 +7,7 @@
 #include <stdint.h>
 #include <stdbool.h>
 #include <rte_common.h>
+#include "dp_util.h"
 
 #ifdef __cplusplus
 extern "C" {
@@ -30,8 +31,8 @@ enum dp_vnf_type {
 } __rte_packed;  // for 'struct dp_flow' and 'struct flow_key'
 
 struct dp_vnf_prefix {
-	uint32_t	ip;
-	uint8_t		length;
+	struct dpgrpc_address ol;
+	uint8_t				  length;
 };
 
 struct dp_vnf {
@@ -45,13 +46,13 @@ int dp_vnf_init(int socket_id);
 void dp_vnf_free(void);
 
 int dp_add_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE], enum dp_vnf_type type,
-			   uint16_t port_id, uint32_t vni, uint32_t prefix_ip, uint8_t prefix_len);
+			   uint16_t port_id, uint32_t vni, struct dpgrpc_address *pfx_ip, uint8_t prefix_len);
 const struct dp_vnf *dp_get_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE]);
 int dp_del_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE]);
 
-bool dp_vnf_lbprefix_exists(uint16_t port_id, uint32_t vni, uint32_t prefix_ip, uint8_t prefix_len);
+bool dp_vnf_lbprefix_exists(uint16_t port_id, uint32_t vni, struct dpgrpc_address *pfx_ip, uint8_t prefix_len);
 
-int dp_del_vnf_by_value(enum dp_vnf_type type, uint16_t port_id, uint32_t vni, uint32_t prefix_ip, uint8_t prefix_len);
+int dp_del_vnf_by_value(enum dp_vnf_type type, uint16_t port_id, uint32_t vni, struct dpgrpc_address *pfx_ip, uint8_t prefix_len);
 
 int dp_list_vnf_alias_prefixes(uint16_t port_id, enum dp_vnf_type type, struct dp_grpc_responder *responder);
 
--- a/include/protocols/dp_icmpv6.h
+++ b/include/protocols/dp_icmpv6.h
@@ -1,0 +1,7 @@
+#ifndef __INCLUDE_DP_DHCPV6_H__
+#define __INCLUDE_DP_DHCPV6_H__
+
+#define	DP_ICMPV6_ECHO_REQUEST	128
+#define	DP_ICMPV6_ECHO_REPLY	129
+
+#endif
--- a/include/protocols/dp_icmpv6.h
+++ b/include/protocols/dp_icmpv6.h
@@ -1,5 +1,5 @@
-#ifndef __INCLUDE_DP_DHCPV6_H__
-#define __INCLUDE_DP_DHCPV6_H__
+#ifndef __INCLUDE_DP_ICMPV6_H__
+#define __INCLUDE_DP_ICMPV6_H__
 
 #define	DP_ICMPV6_ECHO_REQUEST	128
 #define	DP_ICMPV6_ECHO_REPLY	129
--- a/src/dp_cntrack.c
+++ b/src/dp_cntrack.c
@@ -149,6 +149,7 @@ static __rte_always_inline struct flow_value *flow_table_insert_entry(struct flo
 {
 	struct flow_value *flow_val;
 	struct flow_key inverted_key;
+	struct dpgrpc_address pfx_ip;
 
 	flow_val = rte_zmalloc("flow_val", sizeof(struct flow_value), RTE_CACHE_LINE_SIZE);
 	if (!flow_val) {
@@ -161,12 +162,15 @@ static __rte_always_inline struct flow_value *flow_table_insert_entry(struct flo
 	flow_val->timeout_value = flow_timeout;
 	flow_val->created_port_id = port->port_id;
 
+	memset(&pfx_ip, 0, sizeof(pfx_ip));
+	pfx_ip.ip_type = RTE_ETHER_TYPE_IPV4;
+	pfx_ip.ipv4 = key->l3_dst.ip4;
 	/* Target ip of the traffic is an alias prefix of a VM in the same VNI on this dp-service */
 	/* This will be an uni-directional traffic, which does not expect its corresponding reverse traffic */
 	/* Details can be found in https://github.com/ironcore-dev/dpservice/pull/341 */
 	if (offload_mode_enabled
 		&& !port->is_pf
-		&& dp_vnf_lbprefix_exists(DP_VNF_MATCH_ALL_PORT_IDS, key->vni, key->l3_dst.ip4, 32)
+		&& dp_vnf_lbprefix_exists(DP_VNF_MATCH_ALL_PORT_IDS, key->vni, &pfx_ip, 32)
 	)
 		flow_val->nf_info.nat_type = DP_FLOW_LB_TYPE_LOCAL_NEIGH_TRAFFIC;
 	else
--- a/src/dp_cntrack.c
+++ b/src/dp_cntrack.c
@@ -302,8 +302,10 @@ int dp_cntrack_handle(struct rte_mbuf *m, struct dp_flow *df)
 	if (df->l4_type == IPPROTO_TCP && df->vnf_type != DP_VNF_TYPE_LB) {
 		if (df->l3_type == RTE_ETHER_TYPE_IPV4)
 			tcp_hdr = (struct rte_tcp_hdr *)(dp_get_ipv4_hdr(m) + 1);
-		else
+		else if (df->l3_type == RTE_ETHER_TYPE_IPV6)
 			tcp_hdr = (struct rte_tcp_hdr *)(dp_get_ipv6_hdr(m) + 1);
+		else
+			return DP_ERROR;
 		dp_cntrack_tcp_state(flow_val, tcp_hdr);
 		dp_cntrack_set_timeout_tcp_flow(m, flow_val, df);
 	}
--- a/src/dp_cntrack.c
+++ b/src/dp_cntrack.c
@@ -188,7 +188,7 @@ static __rte_always_inline struct flow_value *flow_table_insert_entry(struct flo
 	df->dp_flow_hash = dp_get_conntrack_flow_hash_value(key);
 
 	memset(&inverted_key, 0, sizeof(inverted_key));
-	dp_invert_flow_key(key, &inverted_key);
+	dp_invert_flow_key(key, df->l3_type, &inverted_key);
 	flow_val->flow_key[DP_FLOW_DIR_REPLY] = inverted_key;
 	if (DP_FAILED(dp_add_flow(&inverted_key, flow_val)))
 		goto error_add_inv;
@@ -243,14 +243,14 @@ static __rte_always_inline int dp_get_flow_val(struct rte_mbuf *m, struct dp_flo
 		return ret;
 
 	// highly optimized comparator for a specific key size
-	static_assert(sizeof(struct flow_key) == 42, "struct flow_key changed size");
+	static_assert(sizeof(struct flow_key) == 44, "struct flow_key changed size");
 	if (prev_key
 		&& ((const uint64_t *)curr_key)[0] == ((const uint64_t *)prev_key)[0]
 		&& ((const uint64_t *)curr_key)[1] == ((const uint64_t *)prev_key)[1]
 		&& ((const uint64_t *)curr_key)[2] == ((const uint64_t *)prev_key)[2]
 		&& ((const uint64_t *)curr_key)[3] == ((const uint64_t *)prev_key)[3]
 		&& ((const uint64_t *)curr_key)[4] == ((const uint64_t *)prev_key)[4]
-		&& ((const uint16_t *)curr_key)[20] == ((const uint16_t *)prev_key)[20]
+		&& ((const uint32_t *)curr_key)[20] == ((const uint32_t *)prev_key)[20]
 	) {
 		// flow is the same as it was for the previous packet
 		*p_flow_val = cached_flow_val;
@@ -296,7 +296,10 @@ int dp_cntrack_handle(struct rte_mbuf *m, struct dp_flow *df)
 	flow_val->timestamp = rte_rdtsc();
 
 	if (df->l4_type == IPPROTO_TCP && df->vnf_type != DP_VNF_TYPE_LB) {
-		tcp_hdr = (struct rte_tcp_hdr *)(dp_get_ipv4_hdr(m) + 1);
+		if (df->l3_type == RTE_ETHER_TYPE_IPV4)
+			tcp_hdr = (struct rte_tcp_hdr *)(dp_get_ipv4_hdr(m) + 1);
+		else
+			tcp_hdr = (struct rte_tcp_hdr *)(dp_get_ipv6_hdr(m) + 1);
 		dp_cntrack_tcp_state(flow_val, tcp_hdr);
 		dp_cntrack_set_timeout_tcp_flow(m, flow_val, df);
 	}
--- a/src/dp_cntrack.c
+++ b/src/dp_cntrack.c
@@ -149,7 +149,7 @@ static __rte_always_inline struct flow_value *flow_table_insert_entry(struct flo
 {
 	struct flow_value *flow_val;
 	struct flow_key inverted_key;
-	struct dpgrpc_address pfx_ip;
+	struct dp_ip_address pfx_ip;
 
 	flow_val = rte_zmalloc("flow_val", sizeof(struct flow_value), RTE_CACHE_LINE_SIZE);
 	if (!flow_val) {
--- a/src/dp_cntrack.c
+++ b/src/dp_cntrack.c
@@ -162,7 +162,6 @@ static __rte_always_inline struct flow_value *flow_table_insert_entry(struct flo
 	flow_val->timeout_value = flow_timeout;
 	flow_val->created_port_id = port->port_id;
 
-	memset(&pfx_ip, 0, sizeof(pfx_ip));
 	pfx_ip.ip_type = RTE_ETHER_TYPE_IPV4;
 	pfx_ip.ipv4 = key->l3_dst.ip4;
 	/* Target ip of the traffic is an alias prefix of a VM in the same VNI on this dp-service */
@@ -191,7 +190,11 @@ static __rte_always_inline struct flow_value *flow_table_insert_entry(struct flo
 	// Implicit casting from hash_sig_t to uint32_t!
 	df->dp_flow_hash = dp_get_conntrack_flow_hash_value(key);
 
-	memset(&inverted_key, 0, sizeof(inverted_key));
+	/* Zero out the whole address part of the keys for IPv4 as the stack may have garbage in the unused part */
+	if (df->l3_type == RTE_ETHER_TYPE_IPV4) {
+		memset(inverted_key.l3_dst.ip6, 0, sizeof(inverted_key.l3_dst.ip6));
+		memset(inverted_key.l3_src.ip6, 0, sizeof(inverted_key.l3_src.ip6));
+	}
 	dp_invert_flow_key(key, df->l3_type, &inverted_key);
 	flow_val->flow_key[DP_FLOW_DIR_REPLY] = inverted_key;
 	if (DP_FAILED(dp_add_flow(&inverted_key, flow_val)))
--- a/src/dp_cntrack.c
+++ b/src/dp_cntrack.c
@@ -11,7 +11,7 @@
 #include "rte_flow/dp_rte_flow_helpers.h"
 #include "monitoring/dp_graphtrace.h"
 
-static struct flow_key key_cache[2];
+static struct flow_key key_cache[2] = {0};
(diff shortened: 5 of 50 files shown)


src/dp_nat.c
```diff
@@ -296,6 +297,184 @@ void dp_nat_chg_ip(struct dp_flow *df, struct rte_ipv4_hdr *ipv4_hdr,
...
+	ipv6_hdr = dp_get_ipv6_hdr(m);
+	*dest_ip = *(int *)&ipv6_hdr->dst_addr[12];
+	l4_proto = ipv6_hdr->proto;
+
+	// Adjust the packet data to fit IPv4
+	if (rte_pktmbuf_adj(m, sizeof(struct rte_ipv6_hdr) - sizeof(struct rte_ipv4_hdr)) == NULL)
+		return DP_ERROR;
+
+	// Move the Ethernet header to just before the IPv4 header
+	memmove(rte_pktmbuf_mtod(m, uint8_t *), eth_hdr, sizeof(struct rte_ether_hdr));
```
PlagueCZ, line 341: I actually never thought of this in virtual services (where this function is also being performed).
It makes sense that it works, looking at the DPDK source code, but it does seem strange to "shorten" memory and then access it. Maybe a comment about it being intentional would help?

src/dp_nat.c
```diff
@@ -296,6 +297,184 @@ void dp_nat_chg_ip(struct dp_flow *df, struct rte_ipv4_hdr *ipv4_hdr,
...
+		m->l4_type = IPPROTO_ICMP;
+
+		icmp_hdr = (struct rte_icmp_hdr *)(ipv4_hdr + 1);
+		icmp_hdr->icmp_code = 0;
+		if (icmp_hdr->icmp_type == DP_ICMPV6_ECHO_REQUEST)
+			icmp_hdr->icmp_type = RTE_IP_ICMP_ECHO_REQUEST;
+		else if (icmp_hdr->icmp_type == DP_ICMPV6_ECHO_REPLY)
+			icmp_hdr->icmp_type = RTE_IP_ICMP_ECHO_REPLY;
+		else
+			return 0;
```
PlagueCZ, line 392: This should be DP_OK or DP_ERROR

src/dp_nat.c
```diff
@@ -296,6 +297,184 @@ void dp_nat_chg_ip(struct dp_flow *df, struct rte_ipv4_hdr *ipv4_hdr,
...
+	ipv4_hdr->packet_id = 0;
+	ipv4_hdr->fragment_offset = 0;
+	ipv4_hdr->time_to_live = ipv6_hdr->hop_limits;
+	ipv4_hdr->next_proto_id = l4_proto;
+	ipv4_hdr->src_addr = htonl(nat_ip);
+	ipv4_hdr->dst_addr = *dest_ip;
+	ipv4_hdr->hdr_checksum = 0;
+
+	m->tx_offload = 0;
+	m->packet_type &= ~RTE_PTYPE_L3_MASK;
```
PlagueCZ, line 357: At first glance this seems strange to me, but I would have to check it via `dpservice-dump` which is the only thing that relies on this value to be right. If you have a test at hand can you look if the output still makes sense after this? `dp_virtsvc.c` does this in a more complicated way.

src/dp_nat.c
```diff
@@ -296,6 +297,184 @@ void dp_nat_chg_ip(struct dp_flow *df, struct rte_ipv4_hdr *ipv4_hdr,
...
+		ipv6_hdr->proto = IPPROTO_ICMPV6;
+		icmp_hdr = (struct rte_icmp_hdr *)(ipv6_hdr + 1);
+		icmp_hdr->icmp_code = 0;
+		icmp_hdr->icmp_cksum = 0;
+
+		if (icmp_hdr->icmp_type == RTE_IP_ICMP_ECHO_REQUEST)
+			icmp_hdr->icmp_type = DP_ICMPV6_ECHO_REQUEST;
+		if (icmp_hdr->icmp_type == RTE_IP_ICMP_ECHO_REPLY)
+			icmp_hdr->icmp_type = DP_ICMPV6_ECHO_REPLY;
+		icmp_hdr->icmp_cksum = rte_ipv6_udptcp_cksum(ipv6_hdr, icmp_hdr);
```
PlagueCZ, line 471: This piece of code is similar to the other direction, yet it does not use `else if ` and `else error`. Not sure if intentional

src/dp_vnf.c
```diff
@@ -38,21 +38,29 @@ int dp_add_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE], enum dp_vnf_type t
...
 	vnf = rte_zmalloc("vnf_handle_mapping", sizeof(struct dp_vnf), RTE_CACHE_LINE_SIZE);
 	if (!vnf) {
-		DPS_LOG_WARNING("VNF handle allocation failed", DP_LOG_VNF_TYPE(type), DP_LOG_VNI(vni),
-						DP_LOG_PORTID(port_id), DP_LOG_IPV4(prefix_ip), DP_LOG_PREFLEN(prefix_len));
+		if (pfx_ip->ip_type == RTE_ETHER_TYPE_IPV4)
+			DPS_LOG_WARNING("VNF handle allocation failed", DP_LOG_VNF_TYPE(type), DP_LOG_VNI(vni),
+							DP_LOG_PORTID(port_id), DP_LOG_IPV4(pfx_ip->ipv4), DP_LOG_PREFLEN(prefix_len));
+		else
+			DPS_LOG_WARNING("VNF handle allocation failed", DP_LOG_VNF_TYPE(type), DP_LOG_VNI(vni),
+							DP_LOG_PORTID(port_id), DP_LOG_IPV6(pfx_ip->ipv6), DP_LOG_PREFLEN(prefix_len));
```
PlagueCZ, line 46: I think we could create a `DP_LOG_IP()` that uses a ternary for `ip->ip_type` to encapsulate  logging for `struct dp_ip_address` It will simplify code in multiple places in this PR.

src/dp_vnf.c
```diff
@@ -94,16 +102,17 @@ int dp_del_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE])
...
-											 uint32_t prefix_ip, uint8_t prefix_len)
+											 struct dp_ip_address *pfx_ip, uint8_t prefix_len)
 {
 	return (port_id == DP_VNF_MATCH_ALL_PORT_IDS || vnf->port_id == port_id)
 		&& vnf->vni == vni
-		&& vnf->alias_pfx.ip == prefix_ip
+		&& ((vnf->alias_pfx.ol.ip_type == RTE_ETHER_TYPE_IPV4 && vnf->alias_pfx.ol.ipv4 == pfx_ip->ipv4)
+			|| (vnf->alias_pfx.ol.ip_type == RTE_ETHER_TYPE_IPV6 && rte_rib6_is_equal(vnf->alias_pfx.ol.ipv6, pfx_ip->ipv6)))
 		&& vnf->alias_pfx.length == prefix_len
 		&& vnf->type == type;
```
PlagueCZ, line 112: This comparison seems to not check `pfx_ip->ip_type`, not sure if intentional. I guess if the caller makes sure to only have the right type, then it's an optimization, but would need to check.

src/grpc/dp_grpc_conv.cpp
```diff
@@ -179,8 +192,23 @@ void DpToGrpcInterface(const struct dpgrpc_iface *dp_iface, Interface *grpc_ifac
...
+					((uint64_t)mask[4] << 24) | ((uint64_t)mask[5] << 16) |
+					((uint64_t)mask[6] << 8)  | (uint64_t)mask[7];
+
+	uint64_t low = ((uint64_t)mask[8] << 56) | ((uint64_t)mask[9] << 48) |
+					((uint64_t)mask[10] << 40) | ((uint64_t)mask[11] << 32) |
+					((uint64_t)mask[12] << 24) | ((uint64_t)mask[13] << 16) |
+					((uint64_t)mask[14] << 8)  | (uint64_t)mask[15];
+
+	return __builtin_popcountll(high) + __builtin_popcountll(low);
+}
```
PlagueCZ, line 207: Not sure if it would work without testing, but I would try `high = rte_be_to_cpu(*((uint64_t *)mask)` and equivalent `(mask+8)` for low.

src/grpc/dp_grpc_conv.cpp
```diff
@@ -203,19 +231,34 @@ void DpToGrpcFwrule(const struct dp_fwall_rule *dp_rule, FirewallRule *grpc_rule
...
+		dst_ip->set_ipver(IpVersion::IPV4);
+		dst_ip->set_address(GrpcConv::Ipv4ToStr(dp_rule->dest_ip.ipv4));
+		dst_pfx->set_length(__builtin_popcount(dp_rule->dest_mask.ip4));
+	} else {
+		dst_ip->set_ipver(IpVersion::IPV6);
+		inet_ntop(AF_INET6, dp_rule->dest_ip.ipv6, strbuf, sizeof(strbuf));
+		dst_ip->set_address(strbuf);
+		dst_pfx->set_length(DpIpv6MaskToPrefixLen(dp_rule->dest_mask.ip6));
+	}
+
```
PlagueCZ, line 260: This is the third(?) time I see this block, I think it could be separated as a `GrpcConv::DpToGrpcPrefix` or something

src/grpc/dp_grpc_impl.c
```diff
@@ -59,22 +59,24 @@ static __rte_always_inline void dp_generate_underlay_ipv6(uint8_t route[DP_VNF_I
...
 	struct dpgrpc_lb *request = &responder->request.add_lb;
 	struct dpgrpc_ul_addr *reply = dp_grpc_single_reply(responder);
+	struct dp_ip_address pfx_ip;
 
 	uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE];
 	int ret = DP_GRPC_OK;
 
-	if (request->addr.ip_type == RTE_ETHER_TYPE_IPV4) {
-		if (DP_FAILED(dp_create_vnf_route(ul_addr6, DP_VNF_TYPE_LB, request->vni, dp_get_pf0(), 0, 0))) {
+	memset(&pfx_ip, 0, sizeof(pfx_ip));
```
PlagueCZ, line 77: I think the intention is to create an equivalent of `0.0.0.0`, but this `pfx_ip` is invalid, because in most places in the code you check the `ip_type` to be IPv4 or IPv6 and fail otherwise.

src/grpc/dp_grpc_impl.c
```diff
@@ -393,20 +397,28 @@ static int dp_process_create_prefix(struct dp_grpc_responder *responder)
...
-		rte_memcpy(reply->addr6, ul_addr6, sizeof(reply->addr6));
+	} else if (request->addr.ip_type == RTE_ETHER_TYPE_IPV6) {
+		ret = dp_add_route6(port, iface_vni, 0, request->addr.ipv6, NULL, request->length);
+		if (DP_FAILED(ret))
+			return ret;
+
+		if (DP_FAILED(dp_create_vnf_route(ul_addr6, DP_VNF_TYPE_ALIAS_PFX, iface_vni, port, &request->addr, request->length))) {
+			dp_del_route6(port, iface_vni, request->addr.ipv6, request->length);
+			return DP_GRPC_ERR_VNF_INSERT;
+		}
```
PlagueCZ, line 418: Since you already do that for `dp_create_vnf_route` wouldn't it be beneficial for `dp_add_route` and `dp_del_route` to actually accept `struct dp_ip_address`?
That way this code would be much simpler (and there is the same effect in `dp_process_create_route`.

src/nodes/cls_node.c
```diff
@@ -200,7 +199,7 @@ static __rte_always_inline rte_edge_t get_next_index(__rte_unused struct rte_nod
 			if (is_ipv6_nd(ipv6_hdr))
 				return CLS_NEXT_IPV6_ND;
 			df->l3_type = ntohs(ether_hdr->ether_type);
-			return CLS_NEXT_IPV6_LOOKUP;
+			return CLS_NEXT_CONNTRACK;
```
PlagueCZ, line 202: Please don't tell me the design document for graph nodes needs redrawing again :D

src/nodes/packet_relay_node.c
```diff
@@ -61,9 +86,12 @@ static __rte_always_inline rte_edge_t get_next_index(__rte_unused struct rte_nod
 		return PACKET_RELAY_NEXT_IPIP_ENCAP;
 	}
 
-	if (df->l4_type == DP_IP_PROTO_ICMP)
+	if (df->l4_type == IPPROTO_ICMP)
```
PlagueCZ, line 89: I'm all for this change (again something on my TODO), but it should be then changed everywhere for consistency.

src/nodes/snat_node.c
```diff
@@ -22,37 +23,41 @@ static __rte_always_inline rte_edge_t get_next_index(__rte_unused struct rte_nod
 {
 	struct dp_flow *df = dp_get_flow_ptr(m);
 	struct flow_value *cntrack = df->conntrack;
+	struct snat_data *snat_data_p = NULL;
```
PlagueCZ, line 26: I've only ever seen `*_p` used for double-pointers in this codebase. I think this should stay `snat_data` and the other one can be `snat64_data`

src/nodes/snat_node.c
```diff
@@ -78,22 +83,80 @@ static __rte_always_inline rte_edge_t get_next_index(__rte_unused struct rte_nod
...
 			if (DP_FAILED(dp_add_flow(&cntrack->flow_key[DP_FLOW_DIR_REPLY], cntrack)))
 				return SNAT_NEXT_DROP;
 		}
-		// it is deleted due the case where an init packet/request is sent twice, and no action is done if it is returned here
-		// it has to continue to the follow-up code
-		// return 1;
+
+		if (df->l3_type == RTE_ETHER_TYPE_IPV6 && port->iface.nat_ip && dp_is_ip6_in_nat64_range(df->dst.dst_addr6)
+		    && df->flow_type == DP_FLOW_SOUTH_NORTH) {
+			struct snat_data snat_data = {0};
```
PlagueCZ, line 96: I thought this was a big no-no for me to do, otherwise I would use this quite often I think. If renamed to `snat64_data` as suggested above, I don't think it needs to be here to indicate that it's a "different snat data"

test/test_lb.py
```diff
@@ -122,7 +122,7 @@ def test_nat_to_lb_nat(request, prepare_ipv4, grpc_client, port_redundancy):
 
 	grpc_client.delnat(VM1.name)
 	grpc_client.dellbtarget(lb_name, lb_vm1_ul_ipv6)
-	grpc_client.dellbprefix(VM1.name, lb_ip)
+	grpc_client.dellbprefix(VM1.name, f"{lb_ip}/32")
```
PlagueCZ, line 125: If these functions are now only ever used like this, can you define a `lb_pfx`  constant instead? Or alternatively rename them to `addlbip` and `dellpip` and do the `f{lb_ip}/32` inside?

PlagueCZ (changes requested):
Code looks good, did not find anything major. Compiles and runs in pytest for me too.
I will try running on HW later, but I put this out first so the minor things can be resolved in the meantime.

hack/dp_conf.json
```diff
@@ -47,6 +47,11 @@
       "arg": "IPv4",
       "help": "set the domain name server field in DHCP responses (can be used multiple times)"
     },
+    {
+		"lgopt": "dhcpv6-dns",
+		"arg": "ADDR6",
+		"help": "set the domain name server field in DHCPv6 responses (can be used multiple times)"
+	},
```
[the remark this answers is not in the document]
guvenc: It was probably my editor. Will correct.

include/dp_util.h
```diff
@@ -21,6 +21,7 @@ extern "C" {
 #define DP_IFACE_PXE_MAX_LEN	32
 #define DP_LB_ID_MAX_LEN		64
 #define DP_LB_MAX_PORTS			16
+#define DP_IPV6_ADDR_SIZE		16
```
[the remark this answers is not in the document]
guvenc: Yeah, corrected.

include/dp_util.h
```diff
@@ -38,6 +39,13 @@ extern "C" {
 #define DP_TCP_PKT_FLAG_SYNACK(FLAGS) \
 	(((FLAGS) & (RTE_TCP_SYN_FLAG|RTE_TCP_ACK_FLAG)) == (RTE_TCP_SYN_FLAG|RTE_TCP_ACK_FLAG))
 
+struct dp_ip_address {
+	uint32_t				ip_type;
```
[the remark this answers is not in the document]
guvenc: Well dp_ip_address is just a rename of dp_grpc_address which was convenient to use to hold both ip versions in a single struct and it had its ip_type as uint32_t which kinda stayed here.
uint16_t seems to make more sense.

include/dp_vnf.h
```diff
@@ -45,13 +46,13 @@ int dp_vnf_init(int socket_id);
...
-			   uint16_t port_id, uint32_t vni, uint32_t prefix_ip, uint8_t prefix_len);
+			   uint16_t port_id, uint32_t vni, struct dp_ip_address *pfx_ip, uint8_t prefix_len);
 const struct dp_vnf *dp_get_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE]);
 int dp_del_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE]);
 
-bool dp_vnf_lbprefix_exists(uint16_t port_id, uint32_t vni, uint32_t prefix_ip, uint8_t prefix_len);
+bool dp_vnf_lbprefix_exists(uint16_t port_id, uint32_t vni, struct dp_ip_address *pfx_ip, uint8_t prefix_len);
 
-int dp_del_vnf_by_value(enum dp_vnf_type type, uint16_t port_id, uint32_t vni, uint32_t prefix_ip, uint8_t prefix_len);
+int dp_del_vnf_by_value(enum dp_vnf_type type, uint16_t port_id, uint32_t vni, struct dp_ip_address *pfx_ip, uint8_t prefix_len);
```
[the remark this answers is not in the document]
guvenc: No strong preference. I just tend to use abbreviations for variables. 
It is better to stay conform to current usage though.

include/protocols/dp_icmpv6.h
```diff
@@ -0,0 +1,7 @@
+#ifndef __INCLUDE_DP_DHCPV6_H__
+#define __INCLUDE_DP_DHCPV6_H__
```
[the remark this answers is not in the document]
guvenc: Corrected

src/dp_cntrack.c
```diff
@@ -161,12 +162,15 @@ static __rte_always_inline struct flow_value *flow_table_insert_entry(struct flo
 	flow_val->timeout_value = flow_timeout;
 	flow_val->created_port_id = port->port_id;
 
+	memset(&pfx_ip, 0, sizeof(pfx_ip));
+	pfx_ip.ip_type = RTE_ETHER_TYPE_IPV4;
+	pfx_ip.ipv4 = key->l3_dst.ip4;
```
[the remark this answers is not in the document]
guvenc: Yeah, indeed, memset() can be omitted here but there are places in the code where dp_ip_address struct is part of a hash table key and there can be some garbage in the ipv6 part for an ipv4 key and the key becomes a different one.

src/dp_cntrack.c
```diff
@@ -187,7 +191,8 @@ static __rte_always_inline struct flow_value *flow_table_insert_entry(struct flo
 	// Implicit casting from hash_sig_t to uint32_t!
 	df->dp_flow_hash = dp_get_conntrack_flow_hash_value(key);
 
-	dp_invert_flow_key(key, &inverted_key);
+	memset(&inverted_key, 0, sizeof(inverted_key));
+	dp_invert_flow_key(key, df->l3_type, &inverted_key);
```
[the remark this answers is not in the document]
guvenc: I totally agree. The problem is though that inverted_key is on stack and if not memseted and the traffic is ipv4 then stack's garbage in ipv6 part will go into the hash key.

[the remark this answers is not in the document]
guvenc: Just tried and if I remove the memset, the pytests start to fail as I expected.

src/dp_cntrack.c
```diff
@@ -242,11 +247,14 @@ static __rte_always_inline int dp_get_flow_val(struct rte_mbuf *m, struct dp_flo
...
-	static_assert(sizeof(struct flow_key) == 18, "struct flow_key changed size");
+	static_assert(sizeof(struct flow_key) == 44, "struct flow_key changed size");
 	if (prev_key
 		&& ((const uint64_t *)curr_key)[0] == ((const uint64_t *)prev_key)[0]
 		&& ((const uint64_t *)curr_key)[1] == ((const uint64_t *)prev_key)[1]
-		&& ((const uint16_t *)curr_key)[8] == ((const uint16_t *)prev_key)[8]
+		&& ((const uint64_t *)curr_key)[2] == ((const uint64_t *)prev_key)[2]
+		&& ((const uint64_t *)curr_key)[3] == ((const uint64_t *)prev_key)[3]
+		&& ((const uint64_t *)curr_key)[4] == ((const uint64_t *)prev_key)[4]
+		&& ((const uint32_t *)curr_key)[20] == ((const uint32_t *)prev_key)[20]
```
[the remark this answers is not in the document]
guvenc: I would leave this as refactoring to another PR. Not really connected with this PR.

src/dp_conf.c
```diff
@@ -62,13 +63,44 @@ const struct dp_conf_dhcp_dns *dp_conf_get_dhcp_dns(void)
...
+
+	tmp = (uint8_t *)realloc(dhcpv6_dns.array, dhcpv6_dns.len + 16);
+	if (!tmp) {
+		DP_EARLY_ERR("Cannot allocate memory for DHCPv6 DNS address");
+		return DP_ERROR;
+	}
+	dhcpv6_dns.array = tmp;
+
+	memcpy(&dhcpv6_dns.array[dhcpv6_dns.len], &addr6.s6_addr, sizeof(addr6.s6_addr));
+	dhcpv6_dns.len += 16;
```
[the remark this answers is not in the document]
guvenc: Sure. Makes sense.

src/dp_firewall.c
```diff
@@ -117,8 +128,23 @@ static __rte_always_inline bool dp_is_rule_matching(const struct dp_fwall_rule *
...
+		dp_apply_ipv6_mask(rule->src_ip.ipv6, rule->src_ip.ipv6, masked_r_src_ip6);
+		dp_apply_ipv6_mask(rule->dest_ip.ipv6, rule->dest_ip.ipv6, masked_r_dest_ip6);
+
+		is_ip_match = rte_rib6_is_equal(masked_src_ip6, masked_r_src_ip6) &&
+					rte_rib6_is_equal(masked_dest_ip6, masked_r_dest_ip6);
+	} else {
+		return false;
+	}
+
+	return (is_ip_match &&
```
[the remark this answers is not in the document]
guvenc: Done. Makes sense. I also noticed a bug in the ipv6 masking and fixed that one as well.

src/dp_flow.c
```diff
@@ -124,8 +152,23 @@ int dp_build_flow_key(struct flow_key *key /* out */, struct rte_mbuf *m /* in *
 	const struct dp_port *port = dp_get_in_port(m);
 	int ret = DP_OK;
 
-	key->ip_dst = ntohl(df->dst.dst_addr);
-	key->ip_src = ntohl(df->src.src_addr);
+	memset(key, 0, sizeof(struct flow_key));
```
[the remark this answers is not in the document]
guvenc: Makes sense.

src/dp_lb.c
```diff
@@ -226,25 +248,33 @@ static int dp_lb_rr_backend(struct lb_value *val, const struct lb_port *lb_port)
...
-uint8_t *dp_lb_get_backend_ip(uint32_t ol_ip, uint32_t vni, rte_be16_t port, uint8_t proto)
+uint8_t *dp_lb_get_backend_ip(struct flow_key *fkey, uint32_t vni)
 {
 	struct lb_value *lb_val = NULL;
-	struct lb_key nkey = {
-		.ip = ol_ip,
-		.vni = vni
-	};
 	struct lb_port lb_port;
+	struct lb_key nkey;
```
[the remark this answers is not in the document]
guvenc: Done

src/dp_lb.c
```diff
@@ -226,25 +248,33 @@ static int dp_lb_rr_backend(struct lb_value *val, const struct lb_port *lb_port)
 	return ret;
 }
 
-uint8_t *dp_lb_get_backend_ip(uint32_t ol_ip, uint32_t vni, rte_be16_t port, uint8_t proto)
+uint8_t *dp_lb_get_backend_ip(struct flow_key *fkey, uint32_t vni)
```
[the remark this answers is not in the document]
guvenc: Done

src/dp_lb.c
```diff
@@ -65,10 +65,18 @@ static int dp_map_lb_handle(const void *id_key, const struct lb_key *l_key, stru
...
 	struct lb_value *lb_val = NULL;
-	struct lb_key nkey = {
-		.ip = lb->addr.ipv4,
-		.vni = lb->vni
-	};
+	struct lb_key nkey;
+
+	memset(&nkey, 0, sizeof(nkey));
+	nkey.vni = lb->vni;
+	nkey.ip_type = lb->addr.ip_type;
```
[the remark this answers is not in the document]
guvenc: Makes sense.

src/dp_nat.c
guvenc: Added a comment.

src/dp_nat.c
guvenc: Good catch. Corrected.

src/dp_nat.c
guvenc: Was not intentional. The blocks in both directions should look and behave similar. Corrected.

src/dp_vnf.c
guvenc: You mean smt like this:

> #define DP_LOG_IP(ip_addr) \
       ((ip_addr)->ip_type == RTE_ETHER_TYPE_IPV4 ? \
               DP_LOG_IPV4((ip_addr)->ipv4) : \
               DP_LOG_IPV6((ip_addr)->ipv6))

But ternary would return integer for the ipv4 case and pointer for the other case. I would leave it as it is and we can improve the logging as a refactoring

PlagueCZ: Oh you're right, the ternary is runtime, but the macro is compile-time oriented, sorry.
you need to actually define a new type:
```
#define _DP_LOG_IPVx(KEY, VALUE) KEY, (VALUE)->ip_type == RTE_ETHER_TYPE_IPV4 ? _DP_LOG_FMT_IPV4 : _DP_LOG_FMT_IPV6, VALUE
```
And then the macro itself
```
#define DP_LOG_IP(VALUE) _DP_LOG_IP((VALUE)->ip_type == RTE_ETHER_TYPE_IPV4 ? "ipv4" : "ipv6", (VALUE)->ip_type == RTE_ETHER_TYPE_IPV4 ? (VALUE)->ipv4 : (VALUE)->ipv6)
```
But yea, I can just note it down for later commit in logging.

src/dp_nat.c
guvenc: Ok. When you now say it, I checked the dp_virtsvc.c and the realized that I am missing the line
`m->packet_type |= RTE_PTYPE_L3_IPV4;`
It is even better to combine everything to a single line so I would go with the dp_virtsvc approach.
`m->packet_type = (m->packet_type & ~RTE_PTYPE_L3_MASK) | RTE_PTYPE_L3_IPV4;`

src/dp_vnf.c
guvenc: Adding it would make sense, we may have the case where the lower bytes of the ipv6 address matches to ipv4 comparison :)

src/grpc/dp_grpc_conv.cpp
guvenc: You mean smt like this:

>      uint64_t high = rte_be_to_cpu_64(*((const uint64_t *)(mask)));
>      uint64_t low = rte_be_to_cpu_64(*((const uint64_t *)(mask + 8)));

Seems to work in my single test case so I will convert it to this one as it seems more readable.

src/grpc/dp_grpc_conv.cpp
guvenc: Done.

src/grpc/dp_grpc_impl.c
guvenc: Replaced it with

> 	struct dp_ip_address pfx_ip = {
		.ip_type = request->addr.ip_type
	};

for every occurrence.

src/grpc/dp_grpc_impl.c
guvenc: I would agree but this is more of a refactoring which we can do in a follow-up refactoring PR. This would bloat the diff here as well.

src/nodes/cls_node.c
guvenc: I will provide the updated design document in a follow-up PR.

src/nodes/packet_relay_node.c
guvenc: Yeah I agree but current code base has already mixed usage so this one is not making it worse so let's clean it up in a follow-up refactoring PR.

src/nodes/snat_node.c
guvenc: Your are right. I also ask myself, why I declared it here and not in the beginning of the function and the above name change to _p was also to avoid the shadowing. So fixed both.

test/test_lb.py
guvenc: Done

test/test_pf_to_vf.py
```diff
@@ -26,9 +26,35 @@ def test_pf_to_vf_lb_tcp(prepare_ifaces, grpc_client):
 		f"Wrong packet received (ip: {dst_ip}, dport: {dport})"
 
 	grpc_client.dellbtarget(lb_name, lbpfx_ul_ipv6)
-	grpc_client.dellbprefix(VM1.name, lb_ip)
+	grpc_client.dellbprefix(VM1.name, f"{lb_ip}/32")
```
PlagueCZ, line 29: Two forgotten places that should now use `lb_pfx`

test/test_zzz_grpc.py
```diff
@@ -138,22 +138,55 @@ def test_grpc_add_list_delPfx(prepare_ifaces, grpc_client):
...
 	specs = grpc_client.listlbprefixes(VM2.name)
 	assert myspec in specs, \
 		f"Loadbalancer prefix {lb_prefix} not added properly"
 	# Try to add/delete to/from a machine which doesnt exist
-	grpc_client.expect_error(205).addlbprefix("invalid_name", lb_ip)
-	grpc_client.expect_error(205).dellbprefix("invalid_name", lb_ip)
-	grpc_client.dellbprefix(VM2.name, lb_ip)
+	grpc_client.expect_error(205).addlbprefix("invalid_name", lb_prefix)
+	grpc_client.expect_error(205).dellbprefix("invalid_name", lb_prefix)
+	grpc_client.dellbprefix(VM2.name, lb_prefix)
```
PlagueCZ, line 170: Now that you defined the `lb_pfx`, can you please remove the local `lb_prefix` here? Lines will be changed either way.

test/test_dhcpv6.py
```diff
@@ -27,6 +27,16 @@ def test_dhcpv6(prepare_ifaces):
...
+	# Extracting the DNS servers from the DHCPv6 reply
+	dns_servers = None
+	if DHCP6OptDNSServers in answer:
+		dns_servers = answer[DHCP6OptDNSServers].dnsservers
+	else:
+		raise AssertionError("No DNS servers option in DHCPv6 reply")
+
+	if not dns_servers or dhcpv6_dns1 not in dns_servers or dhcpv6_dns2 not in dns_servers:
+		raise AssertionError(f"DHCPv6 reply does not specify the correct DNS servers: {dns_servers} instead of {dhcpv6_dns1} and {dhcpv6_dns2}")
+
```
PlagueCZ, line 39: Can you change this to use `assert DHCP6OptDNSServers in answer, "No DNS...` and similar for the second condition too?

For one thing, it's more consistent with the rest of the code, for other, AssertionError is an exception, whereas asserts can be actually turned off globally by `-O`

src/grpc/dp_grpc_conv.cpp
```diff
@@ -179,6 +192,34 @@ void DpToGrpcInterface(const struct dpgrpc_iface *dp_iface, Interface *grpc_ifac
 	grpc_iface->set_underlay_route(strbuf);
 }
 
+uint32_t DpIpv6MaskToPrefixLen(const uint8_t *mask)
+{
+	uint64_t high = rte_be_to_cpu_64(*((const uint64_t *)(mask)));
+	uint64_t low = rte_be_to_cpu_64(*((const uint64_t *)(mask + 8)));
+
+
```
PlagueCZ, line 200: Nitpick: double newline here.

src/grpc/dp_grpc_conv.cpp
```diff
@@ -203,19 +245,13 @@ void DpToGrpcFwrule(const struct dp_fwall_rule *dp_rule, FirewallRule *grpc_rule
...
+	SetupIpAndPrefix(dp_rule, src_ip, src_pfx, is_src);
 	grpc_rule->set_allocated_source_prefix(src_pfx);
 
 	dst_ip = new IpAddress();
-	dst_ip->set_ipver(IpVersion::IPV4);
-	dst_ip->set_address(GrpcConv::Ipv4ToStr(dp_rule->dest_ip));
 	dst_pfx = new Prefix();
-	dst_pfx->set_allocated_ip(dst_ip);
-	dst_pfx->set_length(__builtin_popcount(dp_rule->dest_ip_mask));
+	SetupIpAndPrefix(dp_rule, dst_ip, dst_pfx, !is_src);
```
PlagueCZ, line 254: I like that this is more readable than using `true` and `false`, it just tripped me up that it's a variable. Usually I see this done as `#define IS_SRC true` and `#define IS_DST false`, since uppercase does not indicate a variable

src/dp_flow.c
```diff
@@ -148,6 +192,10 @@ int dp_build_flow_key(struct flow_key *key /* out */, struct rte_mbuf *m /* in *
 	case IPPROTO_ICMP:
 		ret = dp_build_icmp_flow_key(df, key, m);
 		break;
+	case IPPROTO_ICMPV6:
+		key->port_dst = ntohs(df->l4_info.icmp_field.icmp_identifier);
+		key->src.type_src = df->l4_info.icmp_field.icmp_type;
+		return DP_OK;
```
PlagueCZ, line 198: This should be a break to make it consistent for all l4 types

src/grpc/dp_grpc_impl.c
```diff
@@ -655,6 +681,9 @@ static int dp_process_delete_nat(struct dp_grpc_responder *responder)
 	reply->addr.ip_type = RTE_ETHER_TYPE_IPV4;
 	reply->addr.ipv4 = s_data->nat_ip;
 	dp_del_vip_from_dnat(s_data->nat_ip, iface_vni);
+	port->iface.nat_ip = 0;
+	port->iface.nat_port_range[0] = 0;
+	port->iface.nat_port_range[1] = 0;
```
PlagueCZ, line 686: Why is this part of the port interface now? Is that because without NAT you cannot support NAT64? (This is just so I can understand what happened, this is not a review comment)

PlagueCZ (changes requested):
Just a few little things I noticed after reading the changes from today.

test/test_pf_to_vf.py
guvenc: Done. There was also a similar usage with "/128" lb_ip6 which I corrected as well.

test/test_zzz_grpc.py
guvenc: Done

test/test_dhcpv6.py
guvenc: This exact same DNS check is actually handled with AssertionError for dhcpv4 case so this was an attempt to stay consistent with the current code base. Will change to asserts. Let's see what we can do with the dhcpv4 AssertionError exceptions later on.

src/grpc/dp_grpc_conv.cpp
guvenc: Corrected.

src/grpc/dp_grpc_conv.cpp
guvenc: I am always skeptical using such #defines in c++ context. Defines are for me too much c-like :)  
I am opting for constexpr then smt like this:

>  constexpr bool IS_SRC = true;
     constexpr bool IS_DST = false;

src/dp_flow.c
guvenc: Yeah but one of the compilers in CI/CD (I think it was clang) with high level warnings enabled was complaining "break" is unreachable code so the break was removed intentionally.

src/grpc/dp_grpc_impl.c
guvenc: So an interface actually can do both NAT (ipv4) and NAT64 (for ipv6 but the NAT address is till ipv4). 
- NAT (ipv4) looks up in the hash table whether its own ipv4 source address supposed to be NATed. This approach stays the same and untouched.
- NAT64 actually decides to do NAT64 only if the target ipv6 address of the packet in a special range NAT64 range AND packet's originating interface has a NAT ip set. (port->iface.nat_ip)

This is to keep the GRPC interface backward compatible and make the NAT64 work side by side to NAT (ipv4). This is the first "backward compatible" version. We can iterate (if needed) and enhance the GRPC interface in the future to support a separate NAT IP for ipv4 and a separate NAT64 ipv4 address for ipv6. 

So `port->iface.nat_ip` is substituting the hash table lookup in SNAT table for NAT64 to answer your question in short.

src/grpc/dp_grpc_conv.cpp
PlagueCZ: Perfect. That is both C++ like and it solves the variable problem. I need to remember this for later use.

src/dp_flow.c
PlagueCZ: Oh, then I guess let's make all cases do a `return` because of that.

src/dp_nat.c
```diff
@@ -296,6 +297,187 @@ void dp_nat_chg_ip(struct dp_flow *df, struct rte_ipv4_hdr *ipv4_hdr,
...
+	ipv4_hdr->next_proto_id = l4_proto;
+	ipv4_hdr->src_addr = htonl(nat_ip);
+	ipv4_hdr->dst_addr = *dest_ip;
+	ipv4_hdr->hdr_checksum = 0;
+
+	m->tx_offload = 0;
+	m->packet_type = (m->packet_type & ~RTE_PTYPE_L3_MASK) | RTE_PTYPE_L3_IPV4;
+	m->ol_flags |= RTE_MBUF_F_TX_IPV4;
+	m->ol_flags |= RTE_MBUF_F_TX_IP_CKSUM;
+	m->tx_offload = 0;
```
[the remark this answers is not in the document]
guvenc: Corrected

src/nodes/snat_node.c
```diff
@@ -22,22 +23,27 @@ static __rte_always_inline rte_edge_t get_next_index(__rte_unused struct rte_nod
 {
```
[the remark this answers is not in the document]
guvenc: Makes sense for better readability. Done.

src/dp_flow.c
guvenc: I corrected it differently.  Take a look after I push the commit.

PlagueCZ (approved):
Looks great now

Issue #456, opened 2023-12-07
Network functionality of metering/rate limiting


byteocean:
The network functionality of metering/rate limiting is currently missing in dpservice. By considering the networking we have, what I understand by browsing the relevant pages from cloud providers is as following:
 1) rate limiting mainly applies onto the egress direction from the VM's perspective. In our setup, performing rate limiting on the ingress direction of VMs by dpservice does not make too much sense, as it cannot actually reduce traffic for the underlay network;
 2) other cloud providers such as GCP have different rate limiting parameters depending on the destinations of traffic flows, e.g, cross-region or per-project basis. As the starting point, in our setup, we could have two rate limiting parameters for each VM. One for public traffic (south-north) and one for private traffic (east-west), and they are both per-VM based. The absolute values of these two parameters can be very different, ranging from dozens of Mbits/s to several Gbits/s.
 
Possible solutions (non-offloading path):
(a) use DPDK's provided metering algorithm and perform egress rate limiting with CPU. [Example](https://github.com/DPDK/dpdk/blob/main/examples/qos_meter/main.c). Cons: maximum flexibility in terms of implementing different rate limiting strategies. Feasible rate limiting strategies include: 1) rate limiting depending on traffic types (public vs. private); 2) rate limiting by bandwidth or pps, which is more crucial for small packets. Pros: it costs CPU cycles to determine which packets should be dropped.
(b) relying on VF tx rate limiting supported by Mellanox hardware (@guvenc ). [Example](https://enterprise-support.nvidia.com/s/article/HowTo-Configure-Rate-Limit-per-VF-for-ConnectX-4-ConnectX-5-ConnectX-6#jive_content_id_sysfs_Option_2). Cons: relatively easy to change on the system, as corresonding sys files of VFs needs to change. And hardware would take care of performing rate limiting on the egress direction. Pros: lack of the fexibility to implement various rate limiting strategies described above. For example, if a bandwidth cap of 1Gbits/s is written to the sys file, both directions, namely north-south and east-west, have to conform this value. Addtionally, pps based rate limiting is not supported.

What we could do here is to combine both approaches to achieve max benefits. For each VF, we could have a default bandwidth cap enforced by the approach (b), e.g., 10 Gbit/s for a normal VM. Or 100 Mbit/s for a small testing VM. For each created interface, via gRPC calls, we futher specify more concrete limiting strategies by (a), such as if using bandwidth-based approach or pps-based approach, and specify different values for public and private traffic.

[...]

PR #463, opened 2023-12-21
Perform traffic metering on vf's tx traffic

byteocean:
Solution for #456.

### DIFF
--- a/src/dp_port.c
+++ b/src/dp_port.c
@@ -22,6 +22,10 @@
 #define DP_PORT_INIT_PF true
 #define DP_PORT_INIT_VF false
 
+#define DP_METER_CIR_BASE_VALUE  1024 * 1024 // 1 Mbits
+#define DP_METER_EBS_BREAK_VALUE 100     // 100 Mbits/s
+#define DP_METER_MBITS_TO_BYTES  (1024 * 1024 / 8)
+
 static const struct rte_eth_conf port_conf_default = {
 	.rxmode = {
 		.mq_mode = RTE_ETH_MQ_RX_NONE,
@@ -42,6 +46,12 @@ static const struct rte_eth_conf port_conf_default = {
 	.intr_conf = {
 		.lsc = 1, /**< lsc interrupt feature enabled */
 	},
+};
+
+static const struct rte_meter_srtcm_params dp_srtcm_params_base = {
+	.cir = DP_METER_CIR_BASE_VALUE * 100 / 8,	// 100 Mbits/s
+	.cbs = 4096,			// 4 KBytes
+	.ebs = DP_METER_EBS_BREAK_VALUE * DP_METER_MBITS_TO_BYTES,			// DP_METER_EBS_BREAK_VALUE -> kbps -> * 1000 -> bytes
 };
 
 struct dp_port *_dp_port_table[DP_MAX_PORTS];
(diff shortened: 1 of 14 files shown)


src/grpc/dp_grpc_impl.c
```diff
@@ -502,6 +502,12 @@ static int dp_process_create_interface(struct dp_grpc_responder *responder)
 		goto route6_err;
 	}
 
+	if (!port->is_pf)
```
guvenc, line 505: You only touch create/delete interface calls. What about get and list interface calls ?

src/grpc/dp_grpc_impl.c
```diff
@@ -502,6 +502,12 @@ static int dp_process_create_interface(struct dp_grpc_responder *responder)
...
+		if (DP_FAILED(dp_port_meter_config(port, request->total_flow_rate_cap, request->public_flow_rate_cap))) {
```
guvenc, line 506: We should also store the total_flow and public_flow rates in the dp_port's dp_port_iface sub structure so the get and list GRPC calls can return what was set. This should eliminate the boolean flag in dp_port, as if public_flow is zero, it is not set and also the same for total_flow. The checks whether these are 0 or not can be also put behind a macro for better readibility.

And the second thing is if total_flow_rate_cap and public_flow_rate_cap are not set, we should not call dp_port_meter_config() at all. This will ensure the backward compatibility and the TAP device compatibility will be also ensured. (The caller should not set the metering parameters and our tests do not set it)

guvenc (changes requested):
Thanks. I have some comments in the PR.

src/dp_port.c
```diff
@@ -458,3 +468,63 @@ int dp_stop_port(struct dp_port *port)
...
+	ret = rte_meter_srtcm_profile_config(&port->port_srtcm_profile, &srtcm_params);
+	if (DP_FAILED(ret)) {
+		DPS_LOG_ERR("Cannot configure meter profile", DP_LOG_PORT(port), DP_LOG_RET(ret));
+		return DP_ERROR;
+	}
+
+	ret = rte_meter_srtcm_config(&port->port_srtcm, &port->port_srtcm_profile);
+	if (DP_FAILED(ret)) {
+		DPS_LOG_ERR("Cannot configure meter", DP_LOG_PORT(port), DP_LOG_RET(ret));
+		return DP_ERROR;
```
PlagueCZ, line 495: Now this may not be necessary (you will know the answer to this), but just for completeness, the error branch should reset the srtcm_profile structure from previous block.

src/dp_util.c
```diff
@@ -160,3 +166,45 @@ void dp_free_jhash_table(struct rte_hash *table)
...
+	uint16_t vf_pattern_len = 0;
+	const char *pattern = dp_conf_get_vf_pattern();
+	FILE *fp;
+	struct dp_port *port = dp_get_port_by_id(port_id);
+	uint64_t rate_in_mbits = rate;
+
+	if (!port) {
+		DPS_LOG_ERR("Cannot get port by id", DP_LOG_PORTID(port_id));
+		return DP_ERROR;
+	}
```
PlagueCZ, line 182: No need to log here (unless you want to show the current function) this already logs the unlikely situation when the port is not registered in dpservice.

src/dp_util.c
```diff
@@ -160,3 +166,45 @@ void dp_free_jhash_table(struct rte_hash *table)
...
+				DP_SYSFS_PREFIX_MLX_DEVICE,
+				dp_conf_get_pf0_name(),
+				DP_SYSFS_PREFIX_MLX_MAX_TX_RATE,
+				port->vf_name + vf_pattern_len,
+				DP_SYSFS_SUFFIX_MLX_MAX_TX_RATE)
+			>= (int)sizeof(filename)
+	) {
+		DPS_LOG_ERR("SR-IOV sysfs path to vf's max tx rate is too long");
+		return DP_ERROR;
+	}
```
PlagueCZ, line 198: The better check is to actually cast `(unsigned)printf()` because there can be a negative result (encoding error) and by casting it to unsigned, you force it to be higher than `sizeof(filename)`

PlagueCZ (changes requested):
Seems pretty nice and simple. I only found a few details.

src/dp_port.c
```diff
@@ -22,6 +22,10 @@
 #define DP_PORT_INIT_PF true
 #define DP_PORT_INIT_VF false
 
+#define DP_METER_CIR_BASE_VALUE  1024 * 1024 // 1 Mbits
```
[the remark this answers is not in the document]
byteocean: added.

src/dp_port.c
```diff
@@ -458,3 +468,63 @@ int dp_stop_port(struct dp_port *port)
...
+		return DP_ERROR;
+	}
+
+	return DP_OK;
+}
+
+int dp_port_meter_config(struct dp_port *port, uint64_t total_flow_rate_cap, uint64_t public_flow_rate_cap)
+{
+	if (public_flow_rate_cap > total_flow_rate_cap) {
+		DPS_LOG_ERR("Public flow rate cap cannot be greater than total flow rate cap", DP_LOG_PORT(port));
```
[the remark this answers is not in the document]
byteocean: added.

src/dp_port.c
byteocean: indeed it was not necessary as it is a value assignment and a decision value will take effect if assighment failed here. But for the completeness, the struct in the previous block is set to 0 to erase the assigned values

src/dp_util.c
byteocean: log info is removed

src/dp_util.c
byteocean: used a more explicit way to check and also adapt another location with the same issue.

src/grpc/dp_grpc_impl.c
byteocean: thanks for the comments. I just summarise the the changes to address the above comments since they are highly related.
1) public_flow_rate and total_flow_rate  are added into dp_port_iface, and thus the get/list interface calls will also include flow rates information (corresponding Go part is also adapted);
2) soft_metering_enabled is removed as suggested due to the change in 1);
3) to handle TAP device in the test scenario, the approach to get current tap type is used inside dp_port_meter_config(), to be consistent with other part of the port config code handling test cases. 

Relying only the value of total_flow_rate_cap and public_flow_rate_cap can work. but I found that it is not sufficient in a common development case, and we still need to do a config on sysfs file.
 ```
./dpservice-cli init
./dpservice-cli create interface --id=vm-1 --vni=100 --ipv4=192.168.129.5 --ipv6=2002::123 --device=8a:00.0_representor_vf0 --total-meter-rate=1000
```
then we finish some tests and stop dpservice, then run it again. This time, we create the interface without any metering parameter.
```
./dpservice-cli create interface --id=vm-1 --vni=100 --ipv4=192.168.129.5 --ipv6=2002::123 --device=8a:00.0_representor_vf0
```
If we don't do a config on sysfs file and reset the value to 0, the configured total-meter-rate would still take effect, which is not nice.

guvenc: @byteocean 
Thanks. The changes for my last comments look good. I also agree with setting to zero, as the TAP device case is now also handled nicely in a different way.

There is still a problem though. You go to `err` label when the error `DP_GRPC_ERR_PORT_METER` occurs without doing any rollback so you left a "broken" interface state behind, if that error occurs. 

And the `dp_start_port `is also already called and it is not so easy to rollback the started port. My suggestion is to place the `dp_port_meter_config `before `dp_start_port `and arrange the error labels in such a way that `dp_start_port `failure can also rollback the `dp_port_meter_config `call.

guvenc (changes requested):

src/grpc/dp_grpc_impl.c
byteocean: nice spot and thanks for the suggestion to change. done. I also looked the counter part in `dp_process_delete_interface`, nothing similar needs to be worried there.

byteocean:
rebased to main.

guvenc (approved):
Looks good now.

Issue #464, opened 2024-01-08
No IA_PD and IA_NA address is reported by certain systemd


byteocean:
Running a VM with kernel 6.0.0-4-amd64 and systemd (252 (252.1-1)) leads to the following systemd debug info, and the result is that it fails to obtain the configured ipv6 address. It may not be a direct issue inside dpservice, but rather a finding of a possible compatibility issue, as other standalone dhcp6 client on the same VM actually works.

```
Jan 08 14:04:57 localhost systemd-networkd[1364]: enp2s0: DHCPv6 client: Sent Solicit
Jan 08 14:04:57 localhost systemd-networkd[1364]: enp2s0: DHCPv6 client: Next retransmission in 2s
Jan 08 14:04:57 localhost systemd-networkd[1364]: enp2s0: DHCPv6 client: Received an IA option without valid IA addresses or PD prefixes, ignoring.
Jan 08 14:04:57 localhost systemd-networkd[1364]: enp2s0: DHCPv6 client: Failed to parse IA_NA option, ignoring: No data available
Jan 08 14:04:57 localhost systemd-networkd[1364]: enp2s0: DHCPv6 client: No IA_PD prefix or IA_NA address received. Ignoring.
Jan 08 14:04:57 localhost systemd-networkd[1364]: enp2s0: DHCPv6 client: Failed to process received Advertise message, ignoring: Invalid argument
Jan 08 14:04:59 localhost systemd-networkd[1364]: enp2s0: DHCPv6 client: Sent Solicit
Jan 08 14:04:59 localhost systemd-networkd[1364]: enp2s0: DHCPv6 client: Next retransmission in 3s
Jan 08 14:04:59 localhost systemd-networkd[1364]: enp2s0: DHCPv6 client: Received an IA option without valid IA addresses or PD prefixes, ignoring.
Jan 08 14:04:59 localhost systemd-networkd[1364]: enp2s0: DHCPv6 client: Failed to parse IA_NA option, ignoring: No data available
Jan 08 14:04:59 localhost systemd-networkd[1364]: enp2s0: DHCPv6 client: No IA_PD prefix or IA_NA address received. Ignoring.
Jan 08 14:04:59 localhost systemd-networkd[1364]: enp2s0: DHCPv6 client: Failed to process received Advertise message, ignoring: Invalid argument
```

byteocean:
closed by #469

PR #469, opened 2024-01-10
Do not pad dhcp6 packets and options

byteocean:
possible padding leads to dhcp6 failure described in #464.


byteocean:
fix #464

guvenc (commented):
The PR looks ok but just for my understanding before I merge. Was it really necessary to remove the packed on each struct ? I mean, apart from `dhcpv6_packet` struct, all other structs seem to be already 2 and 4 byte aligned, so the packed didnt actually do anything.

guvenc (approved):

PR #473, opened 2024-01-15
Enhance GRPC function implementations to support different combination of ipv4/ipv6 address cases

guvenc:
Enhance GRPC function implementations to support different combination of ipv4/ipv6 address cases.
After ipv6 introduction, having only ipv6 address or having only ipv4 addresses (with enabled overall ipv6 flag of dpservice)
on dpservice interfaces are valid cases. These should be handled accordingly in the grpc implementation of 
relevant functions. (GetVNI, CreateInterface)

### DIFF
--- a/include/dp_port.h
+++ b/include/dp_port.h
@@ -39,7 +39,6 @@ struct dp_port_iface {
 	uint8_t					ul_ipv6[DP_IPV6_ADDR_SIZE];
 	uint32_t				nat_ip;
 	uint16_t				nat_port_range[2];
-	bool					is_ipv6_set;
 	bool					ready;
 };
 
--- a/include/dp_port.h
+++ b/include/dp_port.h
@@ -39,6 +39,7 @@ struct dp_port_iface {
 	uint8_t					ul_ipv6[DP_IPV6_ADDR_SIZE];
 	uint32_t				nat_ip;
 	uint16_t				nat_port_range[2];
+	bool					is_ipv6_set;
 	bool					ready;
 };
 
--- a/src/dp_periodic_msg.c
+++ b/src/dp_periodic_msg.c
@@ -15,6 +15,11 @@
 static uint8_t dp_mc_ipv6[16] = { 0xff, 0x02, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x01 };
 static uint8_t dp_mc_mac[6] = { 0x33, 0x33, 0x00, 0x00, 0x00, 0x01 };
 
+static __rte_always_inline bool dp_is_ip_set(struct dp_port *port, uint16_t eth_type)
+{
+	return (eth_type == RTE_ETHER_TYPE_IPV6 && !dp_is_ipv6_addr_zero(port->iface.cfg.dhcp_ipv6)) ||
+		   (eth_type == RTE_ETHER_TYPE_ARP && port->iface.cfg.own_ip != 0);
+}
 
 void send_to_all_vfs(const struct rte_mbuf *pkt, uint16_t eth_type)
 {
@@ -30,10 +35,7 @@ void send_to_all_vfs(const struct rte_mbuf *pkt, uint16_t eth_type)
 		if (port->is_pf || !port->allocated)
 			continue;
 
-		if (eth_type == RTE_ETHER_TYPE_IPV6 && !port->iface.is_ipv6_set)
-			continue;
-
-		if (eth_type == RTE_ETHER_TYPE_IPV4 && port->iface.cfg.own_ip == 0)
+		if (!dp_is_ip_set(port, eth_type))
 			continue;
 
 		clone_buf = rte_pktmbuf_copy(pkt, dp_layer->rte_mempool, 0, UINT32_MAX);
--- a/src/dp_periodic_msg.c
+++ b/src/dp_periodic_msg.c
@@ -33,6 +33,9 @@ void send_to_all_vfs(const struct rte_mbuf *pkt, uint16_t eth_type)
 		if (eth_type == RTE_ETHER_TYPE_IPV6 && !port->iface.is_ipv6_set)
 			continue;
 
+		if (eth_type == RTE_ETHER_TYPE_IPV4 && port->iface.cfg.own_ip == 0)
+			continue;
+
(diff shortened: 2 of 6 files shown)


include/dp_port.h
```diff
@@ -39,6 +39,7 @@ struct dp_port_iface {
 	uint8_t					ul_ipv6[DP_IPV6_ADDR_SIZE];
 	uint32_t				nat_ip;
 	uint16_t				nat_port_range[2];
+	bool					is_ipv6_set;
```
PlagueCZ, line 42: I would suggest `ipv6_set`, since `is_xxx` usually means the whole structure, like `port->is_attached`. But that may be my non-english-speaking background.

src/dp_periodic_msg.c
```diff
@@ -30,6 +30,12 @@ void send_to_all_vfs(const struct rte_mbuf *pkt, uint16_t eth_type)
 		if (port->is_pf || !port->allocated)
 			continue;
 
+		if (eth_type == RTE_ETHER_TYPE_IPV6 && !port->iface.is_ipv6_set)
+			continue;
+
+		if (eth_type == RTE_ETHER_TYPE_IPV4 && port->iface.cfg.own_ip == 0)
+			continue;
```
PlagueCZ, line 37: May not be a real problem, but what about other ETHER_TYPEs? Maybe an inline `dp_is_ip_set(port)` or something that does `return (eth_type == RTE_ETHER_TYPE_IPV6 && port->iface.is_ipv6_set) || eth_type == RTE_ETHER_TYPE_IPV4 && port->iface.cfg.own_ip != 0);` ?

src/grpc/dp_grpc_impl.c
```diff
@@ -25,6 +25,14 @@
 
 static uint32_t pfx_counter = 0;
 
+static __rte_always_inline bool dp_is_ipv6_addr_zero(const uint8_t *addr)
+{
+	for (int i = 0; i < DP_IPV6_ADDR_SIZE; i++)
+		if (addr[i] != 0)
+			return false;
+	return true;
+}
```
PlagueCZ, line 34: May be over-optimizing, since I do not know how often this is called, but maybe for the future it would be better to optimize anyway?
```
assert(DP_IPV6_ADDR_SIZE == 2*sizeof(uint64_t), "something something addr size");
return *((const uint64_t *)addr) == 0 && *((const uint64_t *)(addr+sizeof(uint64_t))) == 0;
```

src/grpc/dp_grpc_impl.c
```diff
@@ -485,6 +495,7 @@ static int dp_process_create_interface(struct dp_grpc_responder *responder)
 	rte_memcpy(port->iface.ul_ipv6, ul_addr6, sizeof(port->iface.ul_ipv6));
 	port->iface.cfg.own_ip = request->ip4_addr;
 	port->iface.cfg.ip_depth = DP_LPM_DHCP_IP_DEPTH;
+	port->iface.is_ipv6_set = !dp_is_ipv6_addr_zero(request->ip6_addr);
```
PlagueCZ, line 498: Now seeing this code, we could actually maybe just drop the whole `is_ipv6_set` and use the function dorectly as the optimized version is two CMPs?

PlagueCZ (changes requested):

src/grpc/dp_grpc_impl.c
guvenc: I liked the idea of optimizing the dp_is_ipv6_zero the way you suggested. I think two CMPs also should be good enough, as the loop checking the bool value will be on the "control" thread side.

src/grpc/dp_grpc_impl.c
guvenc: Sounds good to me.

include/dp_port.h
guvenc: I was also hesitant what to choose, ipv6_set sounds a bit better though after giving it a second thought.

src/dp_periodic_msg.c
guvenc: Good catch. Actually ARP needs also consideration here.

src/dp_periodic_msg.c
```diff
@@ -15,6 +15,11 @@
 static uint8_t dp_mc_ipv6[16] = { 0xff, 0x02, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x01 };
 static uint8_t dp_mc_mac[6] = { 0x33, 0x33, 0x00, 0x00, 0x00, 0x01 };
 
+static __rte_always_inline bool dp_is_ip_set(struct dp_port *port, uint16_t eth_type)
+{
+	return (eth_type == RTE_ETHER_TYPE_IPV6 && !dp_is_ipv6_addr_zero(port->iface.cfg.dhcp_ipv6)) ||
+		   (eth_type == RTE_ETHER_TYPE_ARP && port->iface.cfg.own_ip != 0);
+}
```
PlagueCZ, line 22: I am not sure I understand the condition now (or there is a typo). 
IP is set when type is IPV6 and non zero (sure), OR when type is ARP *and* ipv4 is not zero?

PlagueCZ (changes requested):

src/dp_periodic_msg.c
guvenc: If you check the callers of `send_to_all_vfs()` then you will see that the function is actually called to send ARP and ND packets periodically to "all" VFs. (It is only called for ether type ARP and ether type IPv6) 
In case of ipv4 there is the ARP ether type and in case of IPv6, it is the ND packet which is ether type IPv6.
The previous use of ether type IPv4 was not correct so completely removed.

guvenc: Well, this is the condition which we want to omit in the loop where this is called maybe the function name needs to change to smt like `dp_is_port_ok_to_send()`

PlagueCZ: Can you put the summary of your first explanation into a comment there? I get it now, but without your explanation it is unclear from looking at the function.

guvenc: Done

PlagueCZ (approved):
Looks great

PR #474, opened 2024-01-15
Modify dataplane drawing to reflect ipv6 changes

guvenc:
Modify dataplane drawing to reflect ipv6 changes


PlagueCZ (changes requested):
I am not sure how to put this to text, but I would like to have some changes for more clarity and easier reading of the drawing.
1 - move the IPv6 routing node down and left. This way you can eliminate "wire crodding". But you need to move the other nodes a bit to the right maybe.
2 - (this may not be needed if (1) is also done - the DHCPv6 output line breaks uniformiby doing a zig-zag.  You can actually just point it right "after" ARP if you move Conntrack a bit to the right, which would be nice for more space for other texts.
3 - (biggest one I think). Previously CLS had three clear outputs: ARP, IPv4, IPv6. But now there is a non-deterministic state, where IPv6 is on two ouputs of the node. This may be a simple fix of changing the "IPv6" bottom text to "IPv6 ND or underlay".

Or we can sync on this, or I can simply do a commit on top of this one so I can just show you instead of this blob of text.

guvenc:
Thanks Jay. Feel free to do a commit on top of this one with improvements, if you like. I think it might be easier and faster.

PR #476, opened 2024-01-15
Avoid using iteration to determine existence of a vnf value

### LINKED ISSUE #475: Slow data path for VM2VM on the same hypervisor
It was reported that the data path for VM2VM on the same hypervisor is slow. And I was able to reproduce it during recent tests, and it happens for IPv4 traffic.

byteocean:
fix #475

### DIFF
--- a/include/dp_vnf.h
+++ b/include/dp_vnf.h
@@ -52,7 +52,7 @@ int dp_del_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE]);
 
 bool dp_vnf_lbprefix_exists(uint16_t port_id, uint32_t vni, struct dp_ip_address *prefix_ip, uint8_t prefix_len);
 
-int dp_del_vnf_by_value(enum dp_vnf_type type, uint16_t port_id, uint32_t vni, struct dp_ip_address *prefix_ip, uint8_t prefix_len);
+int dp_del_vnf_by_value(struct dp_vnf *target_vnf);
 
 int dp_list_vnf_alias_prefixes(uint16_t port_id, enum dp_vnf_type type, struct dp_grpc_responder *responder);
 
--- a/src/dp_vnf.c
+++ b/src/dp_vnf.c
@@ -11,6 +11,7 @@
 #define DP_VNF_MAX_TABLE_SIZE 1000
 
 static struct rte_hash *vnf_handle_tbl = NULL;
+static struct rte_hash *vnf_value_tbl = NULL;
 
 int dp_vnf_init(int socket_id)
 {
@@ -19,12 +20,59 @@ int dp_vnf_init(int socket_id)
 	if (!vnf_handle_tbl)
 		return DP_ERROR;
 
+	vnf_value_tbl = dp_create_jhash_table(DP_VNF_MAX_TABLE_SIZE, sizeof(struct dp_vnf),
+										  "vnf_value_table", socket_id);
+	if (!vnf_value_tbl) {
+		dp_free_jhash_table(vnf_handle_tbl); // rollback
+		return DP_ERROR;
+	}
+
 	return DP_OK;
 }
 
 void dp_vnf_free(void)
 {
 	dp_free_jhash_table(vnf_handle_tbl);
+}
+
+static int dp_add_vnf_value(struct dp_vnf *vnf)
+{
+	int ret;
+
+	ret = rte_hash_add_key(vnf_value_tbl, vnf);
+	if (DP_FAILED(ret)) {
+		DPS_LOG_ERR("Cannot add VNF value to table", DP_LOG_RET(ret));
+		return DP_ERROR;
+	}
+
+	return DP_OK;
+}
+
+static int dp_delete_vnf_value(struct dp_vnf *vnf)
+{
+	int ret;
+
+	ret = rte_hash_del_key(vnf_value_tbl, vnf);
+	if (DP_FAILED(ret)) {
+		DPS_LOG_ERR("Cannot delete VNF value from table", DP_LOG_RET(ret));
+		return DP_ERROR;
+	}
+
+	return DP_OK;
+}
+
+static bool dp_vnf_value_exists(struct dp_vnf *vnf)
+{
+	int ret;
+
+	ret = rte_hash_lookup(vnf_value_tbl, vnf);
+	if (DP_FAILED(ret)) {
+		if (ret != -ENOENT)
+			DPS_LOG_ERR("Cannot lookup VNF value in table", DP_LOG_RET(ret));
+		return false;
+	}
+
+	return true;
 }
 
 int dp_add_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE], enum dp_vnf_type type,
@@ -65,6 +113,17 @@ int dp_add_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE], enum dp_vnf_type t
 		return DP_ERROR;
 	}
 
+	if (DP_FAILED(dp_add_vnf_value(vnf))) {
+		if (prefix->ip_type == RTE_ETHER_TYPE_IPV4)
+			DPS_LOG_WARNING("Adding VNF value failed", DP_LOG_VNF_TYPE(type), DP_LOG_VNI(vni),
+							DP_LOG_PORTID(port_id), DP_LOG_IPV4(prefix->ipv4), DP_LOG_PREFLEN(prefix_len));
+		else
+			DPS_LOG_WARNING("Adding VNF value failed", DP_LOG_VNF_TYPE(type), DP_LOG_VNI(vni),
+							DP_LOG_PORTID(port_id), DP_LOG_IPV6(prefix->ipv6), DP_LOG_PREFLEN(prefix_len));
+		dp_del_vnf(ul_addr6); //rollback
+		return DP_ERROR;
+	}
+
 	return DP_OK;
 }
 
@@ -89,7 +148,11 @@ int dp_del_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE])
 		return ret;
 	}
 
-	rte_free(vnf);
+	ret = dp_delete_vnf_value(vnf);
+	if (!DP_FAILED(ret)) {
+		DPGRPC_LOG_WARNING("Cannot delete VNF value record");
+		return ret;
+	}
 
 	ret = rte_hash_del_key(vnf_handle_tbl, ul_addr6);
 	if (DP_FAILED(ret)) {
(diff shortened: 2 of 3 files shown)


src/dp_vnf.c
```diff
@@ -27,6 +35,46 @@ void dp_vnf_free(void)
...
+	ret = rte_hash_del_key(vnf_value_tbl, vnf);
+	if (DP_FAILED(ret)) {
+		DPS_LOG_ERR("Cannot delete VNF value from table", DP_LOG_RET(ret));
+		return DP_ERROR;
+	}
+
+	return DP_OK;
+}
+
+static bool dp_if_vnf_value_exists(struct dp_vnf *vnf)
```
PlagueCZ, line 64: `if (dp_if_...` seems strange, `if (dp_vnf_value_exists())` sounds fine

src/dp_vnf.c
```diff
@@ -27,6 +35,46 @@ void dp_vnf_free(void)
 	dp_free_jhash_table(vnf_handle_tbl);
 }
 
+static int dp_record_vnf_value(struct dp_vnf *vnf)
```
PlagueCZ, line 38: This just may be my Slavic mind, but `record` sounds like a world record or music record and not "save" or "add" which this function does :)

src/dp_vnf.c
```diff
@@ -65,6 +113,18 @@ int dp_add_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE], enum dp_vnf_type t
...
 	}
 
+	if (DP_FAILED(dp_record_vnf_value(vnf))) {
+		if (prefix->ip_type == RTE_ETHER_TYPE_IPV4)
+			DPS_LOG_WARNING("VNF value recording failed", DP_LOG_VNF_TYPE(type), DP_LOG_VNI(vni),
+							DP_LOG_PORTID(port_id), DP_LOG_IPV4(prefix->ipv4), DP_LOG_PREFLEN(prefix_len));
+		else
+			DPS_LOG_WARNING("VNF value recording failed", DP_LOG_VNF_TYPE(type), DP_LOG_VNI(vni),
+							DP_LOG_PORTID(port_id), DP_LOG_IPV6(prefix->ipv6), DP_LOG_PREFLEN(prefix_len));
+		dp_delete_vnf_value(vnf);
```
PlagueCZ, line 123: If saving the value failed, then you should not roll it back here, but in the function that failed. Here it has no real effect of course, but in general, that is the pattern everywhere else.

src/dp_vnf.c
```diff
@@ -89,6 +149,12 @@ int dp_del_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE])
 		return ret;
 	}
 
+	ret = dp_delete_vnf_value(vnf);
+	if (!DP_FAILED(ret)) {
+		DPGRPC_LOG_WARNING("Cannot delete VNF value record");
```
PlagueCZ, line 154: Logging twice can sometimes be beneficial, so this is not a change request, just that you could maybe just log here (where there is more information) and simplify the wrapper for delete/record.

src/dp_vnf.c
```diff
@@ -115,25 +181,16 @@ static __rte_always_inline bool dp_vnf_match(const struct dp_vnf *vnf, enum dp_v
...
-		if (dp_vnf_match(vnf, DP_VNF_TYPE_LB_ALIAS_PFX, port_id, vni, prefix_ip, prefix_len))
-			return true;
-	}
-	return false;
+	vnf.port_id = port_id;
+	vnf.vni = vni;
+	vnf.type = DP_VNF_TYPE_LB_ALIAS_PFX;
+	vnf.alias_pfx.length = prefix_len;
+	vnf.alias_pfx.ol.ip_type = prefix_ip->ip_type;
+	memcpy(&vnf.alias_pfx.ol, prefix_ip, sizeof(struct dp_ip_address));
```
PlagueCZ, line 191: You should take advantage of the initialization code to prevent unneeded zeroing and then re-setting:
```
struct dp_vnf vnf = {
    .port_id = port_id,
    ... etc.
}
```
I *think* (but with relative confidence) that even `    .alias_pfx.ol = prefix_ip,` should work jsut fine as it is a shallow copy (even whough of an array).

PlagueCZ (changes requested):

guvenc:
@byteocean 
Automated tests fail. Could you take a look and fix that before we further comment the code ? Thanks.

src/dp_vnf.c
byteocean: changed.

src/dp_vnf.c
byteocean: ja, I was hesitating to use record as well. so changed to "add'

src/dp_vnf.c
byteocean: it was wrong auto-completed code. It should be rolling back the above step.

src/dp_vnf.c
byteocean: yeah, I changed to use the init form. and it works by judging from the test results.

byteocean:
> @byteocean Automated tests fail. Could you take a look and fix that before we further comment the code ? Thanks.

it should be fixed now. the current code is also with changes following existing comments.

src/dp_vnf.c
```diff
@@ -89,79 +148,78 @@ int dp_del_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE])
...
 {
 	struct dp_vnf *vnf;
 	uint32_t iter = 0;
 	int32_t ret;
 	const void *key;
 	int delete_count = 0;
 
+	if (DP_FAILED(dp_delete_vnf_value(target_vnf)))
+		return DP_GRPC_ERR_NOT_FOUND;
+
```
PlagueCZ, line 202: Could you actually improve this further by storing the VNF key (IPv6) as data inside this new table? That way you can remove the (already marked unsafe) hash table traversal below and directly call `dp_del_vnf()` using that key.

I think you only need to change to `static int dp_add_vnf_value(struct dp_vnf *vnf, const uint8_t *ul_addr6)` I think.

This would also eliminate `dp_vnf_match()` from the looks of things.

PlagueCZ (changes requested):
I don't see any problems now, but one thing is possible to do now and I think it should be part of this change.

src/dp_vnf.c
byteocean: the code is optimised to use newly added mapping, but as a separate commit.

include/dp_vnf.h
```diff
@@ -17,6 +17,14 @@ extern "C" {
...
 #define DP_VNF_MATCH_ALL_PORT_IDS 0xFFFF
 
+#define DP_PRINT_VNF_OP_WARNING(MESSAGE, VNF) \
+		if (VNF->alias_pfx.ol.ip_type == RTE_ETHER_TYPE_IPV4) {\
+			DPS_LOG_WARNING(MESSAGE, DP_LOG_VNF_TYPE(VNF->type), DP_LOG_VNI(VNF->vni), \
+							DP_LOG_PORTID(VNF->port_id), DP_LOG_IPV4(VNF->alias_pfx.ol.ipv4), DP_LOG_PREFLEN(VNF->alias_pfx.length)); } \
+		else {\
+			DPS_LOG_WARNING(MESSAGE, DP_LOG_VNF_TYPE(VNF->type), DP_LOG_VNI(VNF->vni), \
+							DP_LOG_PORTID(VNF->port_id), DP_LOG_IPV6(VNF->alias_pfx.ol.ipv6), DP_LOG_PREFLEN(VNF->alias_pfx.length)); } \
+
```
PlagueCZ, line 27: I like the encapsulation for more informative logs.
But, unless you aim to use this macro from another module, it would be best to keep in in the `.c` file. It also can be an inline instead of a macro.
And the name would better be `DP_LOG_VNF_WARNING`, as `PRINT` evokes `printf()`. Not sure what the `_OP_` means, I guess a typo of IP? But even then it is used even for non-IP related warnings, so that's why I would remove that infix also.

src/dp_vnf.c
```diff
@@ -89,83 +156,52 @@ int dp_del_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE])
...
 		return ret;
 	}
 
+	ret = dp_delete_vnf_value(vnf);
+	if (DP_FAILED(ret)) {
+		DPGRPC_LOG_WARNING("Cannot delete VNF value record");
+		return ret;
+	}
+
+	rte_free(vnf);
```
PlagueCZ, line 171: I think you also need to free this when `dp_delete_vnf_value()` fails, because the key is already gone from the `vnf_handle_tbl` thus making this an orphan pointer.

src/dp_vnf.c
```diff
@@ -89,83 +156,52 @@ int dp_del_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE])
...
-					DPS_LOG_ERR("Cannot delete VNF key", DP_LOG_RET(ret), DP_LOG_VNF_TYPE(type), DP_LOG_PORTID(port_id),
-								DP_LOG_VNI(vni), DP_LOG_IPV4(prefix_ip->ipv4), DP_LOG_PREFLEN(prefix_len));
-				else
-					DPS_LOG_ERR("Cannot delete VNF key", DP_LOG_RET(ret), DP_LOG_VNF_TYPE(type), DP_LOG_PORTID(port_id),
-								DP_LOG_VNI(vni), DP_LOG_IPV6(prefix_ip->ipv6), DP_LOG_PREFLEN(prefix_len));
-			}
-		}
-	}
-	return delete_count > 0 ? DP_GRPC_OK : DP_GRPC_ERR_NOT_FOUND;
+	return DP_OK;
```
PlagueCZ, line 204: You can actually just `return dp_del_vnf(vnf_ul_addr6);`

But saying that, this will in the end do another lookup into the `vnf_value_tbl`. So maybe splitting the `dp_del_vnf()` into `static dp_del_vnf_only()` that gets called here is in order (and then deleting from `vnf_value_tbl` explicitely here.

PlagueCZ (changes requested):

include/dp_vnf.h
byteocean: I agree with the critics on including PRINT. The OP means operation, not a typo.

src/dp_vnf.c
byteocean: done.

src/dp_vnf.c
byteocean: it is really a refactoring effort here. It was a tricky to refactor as you suggested (splitting), since this function is heavily used (e.g., in grpc part). I would suggest to address as a separate PR.

byteocean:
rebased to main.

src/dp_vnf.c
```diff
@@ -27,6 +43,64 @@ void dp_vnf_free(void)
...
+		DPS_LOG_WARNING("VNF value allocation failed", DP_LOG_IPV6(ul_addr6));
+		return DP_ERROR;
+	}
+
+	memcpy(vnf_ul_addr6, ul_addr6, DP_IPV6_ADDR_SIZE);
+
+	ret = rte_hash_add_key_data(vnf_value_tbl, vnf, vnf_ul_addr6);
+	if (DP_FAILED(ret)) {
+		DPS_LOG_ERR("Cannot add VNF value and the corresponding underlying IPv6 address to table", DP_LOG_RET(ret));
+		return DP_ERROR;
```
PlagueCZ, line 62: I am not sure how I missed this earlier, but you need to `rte_free(vnf_ul_addr6)` here

src/dp_vnf.c
PlagueCZ: You can make it neater if you free right before the DP_FAILED condition.

PlagueCZ (changes requested):

src/dp_vnf.c
byteocean: Indeed, done

PlagueCZ (approved):

guvenc (approved):
Looks good to me.

PR #479, opened 2024-01-19
Refactor: Use standard protocol defines and remove the self defined ones

guvenc:
Use standard protocol defines and remove the self defined ones


PlagueCZ (approved):
Nice change! I was eyeing this in the past, but thought it was only bugging me and left it for later.

PR #480, opened 2024-01-19
Code fixing to avoid random bits in struct vnf used as key

byteocean:
The bugs were discovered during running dpservice-go tests. 

1) the union part from dp_vnf.alias_pfx needs to pre-filled to avoid random bits in unused places. It is especially important when it is used as hash table keys;
2) the return values from the function `dp_del_vnf_by_value` needs to be consistent with grpc's expected behaviour.

The tests from dpservice passed.
The tests from dpservice-go passed.

### DIFF
--- a/include/dp_vnf.h
+++ b/include/dp_vnf.h
@@ -40,7 +40,7 @@ struct dp_vnf {
 	uint32_t				vni;
 	uint16_t				port_id;
 	struct dp_vnf_prefix	alias_pfx;
-};
+} __rte_packed;
 
 int dp_vnf_init(int socket_id);
 void dp_vnf_free(void);
--- a/include/dp_vnf.h
+++ b/include/dp_vnf.h
@@ -40,6 +40,17 @@ struct dp_vnf {
 	uint32_t				vni;
 	uint16_t				port_id;
 	struct dp_vnf_prefix	alias_pfx;
+};
+
+struct dp_vnf_value_key {
+	enum dp_vnf_type		type;
+	uint32_t				vni;
+	uint16_t				port_id;
+	uint16_t				ip_type;
+	union {
+		uint32_t			ipv4;
+		uint8_t				ipv6[DP_IPV6_ADDR_SIZE];
+	};
 } __rte_packed;
 
 int dp_vnf_init(int socket_id);
--- a/src/dp_vnf.c
+++ b/src/dp_vnf.c
@@ -190,8 +190,10 @@ bool dp_vnf_lbprefix_exists(uint16_t port_id, uint32_t vni, struct dp_ip_address
 		.vni = vni,
 		.type = DP_VNF_TYPE_LB_ALIAS_PFX,
 		.alias_pfx.length = prefix_len,
-		.alias_pfx.ol = *prefix_ip,
+		.alias_pfx.ol = {0},
 	};
+	
+	vnf.alias_pfx.ol = *prefix_ip;
 
 	return dp_vnf_value_exists(&vnf);
 }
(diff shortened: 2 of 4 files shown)


src/dp_vnf.c
```diff
@@ -189,10 +191,12 @@ bool dp_vnf_lbprefix_exists(uint16_t port_id, uint32_t vni, struct dp_ip_address
 		.port_id = port_id,
 		.vni = vni,
 		.type = DP_VNF_TYPE_LB_ALIAS_PFX,
-		.alias_pfx.length = prefix_len,
-		.alias_pfx.ol = *prefix_ip,
 	};
 
+	memset(&vnf.alias_pfx, 0, sizeof(vnf.alias_pfx));
+	vnf.alias_pfx.ol = *prefix_ip;
+	vnf.alias_pfx.length = prefix_len;
```
PlagueCZ, line 198: Wait you are telling me that the old code was not setting this to zero? Because then the C standard is either wrong or I am reading it wrongly and having this problem in the whole codebase.
The old code should behave just like a global variable initialization to zero.

src/grpc/dp_grpc_impl.c
```diff
@@ -391,6 +391,8 @@ static int dp_process_delete_lbprefix(struct dp_grpc_responder *responder)
 
 	target_vnf.port_id = port->port_id;
 	target_vnf.vni = port->iface.vni;
+	target_vnf.alias_pfx.length = request->length;
+	target_vnf.alias_pfx.ol = request->addr;
```
PlagueCZ, line 395: This is the same as above.

src/dp_vnf.c
```diff
@@ -125,6 +125,8 @@ int dp_add_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE], enum dp_vnf_type t
 	vnf->type = type;
 	vnf->vni = vni;
 	vnf->port_id = port_id;
+
+	memset(&vnf->alias_pfx, 0, sizeof(vnf->alias_pfx));
```
PlagueCZ, line 129: You can just do what you are doing everywhere else, i.e. direct initialization using `struct dp_vnf = {}` which should initialize to zero

PlagueCZ (changes requested):

src/dp_vnf.c
byteocean: The tests from dpservice passed.
The tests from dpservice-go passed.

include/dp_vnf.h
```diff
@@ -42,6 +42,17 @@ struct dp_vnf {
 	struct dp_vnf_prefix	alias_pfx;
 };
 
+struct dp_vnf_value_key {
```
guvenc, line 45: @byteocean 
If this is supposed to be reverse lookup of the original hashtable, why do we  redefine a dp_vnf as dp_vnf_value_key ?
dp_vnf can be used as key for reverse lookup ? I need some more background why this needs to be this way.

guvenc: Why cant we just put a packed to the original structure ?

byteocean: yeah, we need to use a data structure without any padding bits to serve as the table key. Although we can make `struct dp_vnf` packed, but we cannot make its nested member `struct dp_ip_address` packed.

guvenc: Ok. Just talked to Jay and he also explained me why.

PlagueCZ (changes requested):

guvenc (commented):
@byteocean 
I have an alternative solution PR  #481 which doesnt use any new structs and transformed keys for reverse lookup. I find it cleaner and easier to tackle the bug. I used also some part of this PR for the error return code fix. Please check.

byteocean:
resolved via #481

PR #481, opened 2024-01-22
Fix VNF reverse lookup

guvenc:
Fix VNF reverse lookup so that the value used as key doesnt contain garbage bytes because of the used union structure for the ip addresses.

### DIFF
--- a/src/dp_vnf.c
+++ b/src/dp_vnf.c
@@ -125,7 +125,7 @@ int dp_add_vnf(const uint8_t ul_addr6[DP_VNF_IPV6_ADDR_SIZE], enum dp_vnf_type t
 	vnf->type = type;
 	vnf->vni = vni;
 	vnf->port_id = port_id;
-	vnf->alias_pfx.ol = *prefix;
+	dp_assign_ip_address(&vnf->alias_pfx.ol, prefix);
 	vnf->alias_pfx.length = prefix_len;
 
 	ret = rte_hash_add_key_data(vnf_handle_tbl, ul_addr6, vnf);
@@ -190,9 +190,9 @@ bool dp_vnf_lbprefix_exists(uint16_t port_id, uint32_t vni, struct dp_ip_address
 		.vni = vni,
 		.type = DP_VNF_TYPE_LB_ALIAS_PFX,
 		.alias_pfx.length = prefix_len,
-		.alias_pfx.ol = *prefix_ip,
 	};
 
+	dp_assign_ip_address(&vnf.alias_pfx.ol, prefix_ip);
 	return dp_vnf_value_exists(&vnf);
 }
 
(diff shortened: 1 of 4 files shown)


guvenc:
@byteocean 
It is not padding that we try to fix here or is this a terminology confusion ? 
It is the union structure with ipv4 and ipv6 address reserving always the size of ipv6 address even for ipv4 address. 
So It "can" contain garbage bytes when coming from the GRPC side. (e.g ipv4 address set and rest of the unused bytes for ipv6 part contain garbage bytes. At least what you guys told me that this was the case)

byteocean:
@guvenc that was not the intention of my comment. It is fully understood what this PR is doing, and it is much similar to the initial solution last week. but the possible padding issue was brought up into the discussion, which leads to the solution of using another data structure with __rte_packed key word. The last comment from me was trying to say, such fix as in this PR should be sufficient, and padding bits should not be the concern.

guvenc:
@byteocean 
Ah ok. There were "two" things the other PR was trying to solve. "padding" and "union" structure garbage bytes.

byteocean:
indeed.

byteocean (approved):

PlagueCZ:
This is what I had in mind when talking about fixing the "address structure problem". So the idea of having an assignment function is a good one, but I think we would have to do it everywhere, not just for VNF, where it happened to cause a visible bug, there are many places where the same bug can still occur. So if this was doe globally for the `dp_ip_address`, then that would be good. But then it would require careful reviews as nothing prevents you from just assigning it manually. 

And of course padding is still a problem when a struct is being used as a key.

src/grpc/dp_grpc_impl.c
```diff
@@ -462,7 +453,7 @@ static int dp_process_delete_prefix(struct dp_grpc_responder *responder)
 	target_vnf.vni = port->iface.vni;
 
 	dp_assign_ip_address(&target_vnf.alias_pfx.ol, &request->addr);
```
PlagueCZ, line 455: Is this needed as an exported thing now? I think you can just assign it here and then do the check in the "key" creation.

I think it would be better because the caller cannot forget to use this special assignment, otherwise stuff will break.

PlagueCZ (changes requested):

src/grpc/dp_grpc_impl.c
byteocean: after doubling checking, this function is exported due to some leftovers. One further step is made to integrate this function into `dp_fill_vnf_data`, as it makes most of the juice from two and reduces code complexity. Unless `dp_assign_ip_address` is needed in other places, the changes now should resolve our concerns so far.

src/dp_vnf.c
```diff
@@ -43,6 +43,24 @@ void dp_vnf_free(void)
...
+	vnf->type = type;
+	vnf->port_id = port_id;
+	vnf->vni = vni;
+	vnf->alias_pfx.length = prefix_len;
+	if (src->ip_type == RTE_ETHER_TYPE_IPV4) {
+		vnf->alias_pfx.ol.ip_type = src->ip_type;
+		vnf->alias_pfx.ol.ipv4 = src->ipv4;
+		return;
+	}
+	vnf->alias_pfx.ol = *src;
```
PlagueCZ, line 61: I know this seems like a nitpick, but in this case I would rather use `if` and `else` as those branches are equal (one is not an error branch to warrant premature return).

PlagueCZ (changes requested):
This seems the best solution to this problem.

src/dp_vnf.c
byteocean: sure, done.

PlagueCZ (approved):

PR #482, opened 2024-01-22
Improve codebase to work with ARM architecture 

guvenc:
- Place iface_id to 4 byte aligned boundary in dpgrpc_prefix
- Setup the interface completely before sending any packet for testing

### DIFF
--- a/include/grpc/dp_grpc_api.h
+++ b/include/grpc/dp_grpc_api.h
@@ -92,9 +92,9 @@ struct dpgrpc_iface_id {
 };
 
 struct dpgrpc_prefix {
+	char					iface_id[DP_IFACE_ID_MAX_LEN];
 	struct dp_ip_address	addr;
 	uint8_t					length;
-	char					iface_id[DP_IFACE_ID_MAX_LEN];
 };
 
 struct dpgrpc_route {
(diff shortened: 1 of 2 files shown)


byteocean:
looks nice.

byteocean (commented):
Would you also provide some background to such changes? does it fail to execute on ARM due to these insufficient parts? thus we could keep it in mind in future development.

guvenc:
> Would you also provide some background to such changes? does it fail to execute on ARM due to these insufficient parts? thus we could keep it in mind in future development.

More background in the commit description:

https://github.com/ironcore-dev/dpservice/pull/482/commits/90da2f990e41c2733e14256091cbbd750446e36b

byteocean (approved):

byteocean:
thanks for pointing to the background of this change.

guvenc:
@PlagueCZ 
The struct looked like this before:

> struct dpgrpc_prefix {
	struct dp_ip_address	addr;
	uint8_t					length;
	char					iface_id[DP_IFACE_ID_MAX_LEN];
};

iface_id is a byte array used in rte_jhash_32b()* for hash calculation. On ARM (maybe it is the default settings of gcc I used there)
iface_id was placed right after uint8_t single byte. (So not on a 4 byte aligned boundary) I guess the struct needs to start memory aligned but not he contents.
On x86 (I didnt check it in debugger but I assume that) it is placed on a 4 byte boundary so we didnt notice anything during hash calculation.

The hash function is doing these kind of dereferencing on the pointer it receives:
`uint32_t val = *((uint32_t*) data)`

PlagueCZ:
> The hash function is doing these kind of dereferencing on the pointer it receives: `uint32_t val = *((uint32_t*) data)`

Ah, thanks this explains it, the requirements are type-dependent (due to how instructions work there), so the cast to uint32 raised the alignment requirement from 1 to 4.

PlagueCZ (approved):
Looks good. Since compiler complains, there is not need for guarding it by reviewers.

PR #485, opened 2024-01-29
Introduce a new NIC type Bluefield2

guvenc:
The sysfs structure on Bluefield2 is different and the VF sysfs entries are actually on the host side and not on the Bluefield side.
React on Bluefield2 card correctly by introducing it as a new NIC type.

### DIFF
--- a/include/dp_conf_opts.h
+++ b/include/dp_conf_opts.h
@@ -9,8 +9,9 @@
 /***********************************************************************/
 
 enum dp_conf_nic_type {
-	DP_CONF_NIC_TYPE_HW,
+	DP_CONF_NIC_TYPE_MELLANOX,
 	DP_CONF_NIC_TYPE_TAP,
+	DP_CONF_NIC_TYPE_BLUEFIELD2,
 };
 
 enum dp_conf_color {
--- a/src/dp_port.c
+++ b/src/dp_port.c
@@ -506,8 +506,12 @@ static int dp_port_public_flow_meter_config(struct dp_port *port, uint64_t publi
 
 int dp_port_meter_config(struct dp_port *port, uint64_t total_flow_rate_cap, uint64_t public_flow_rate_cap)
 {
-	if (dp_conf_get_nic_type() == DP_CONF_NIC_TYPE_TAP)
+	if (dp_conf_get_nic_type() != DP_CONF_NIC_TYPE_MELLANOX) {
+		if (public_flow_rate_cap != 0 || total_flow_rate_cap != 0)
+			DPS_LOG_WARNING("Metering config will not take effect due to the NIC type",
+							DP_LOG_PORT(port), DP_LOG_METER_TOTAL(total_flow_rate_cap), DP_LOG_METER_PUBLIC(public_flow_rate_cap));
 		return DP_OK;
+	}
 
 	if (public_flow_rate_cap > total_flow_rate_cap) {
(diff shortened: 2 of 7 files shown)


src/dp_port.c
```diff
@@ -506,7 +506,7 @@ static int dp_port_public_flow_meter_config(struct dp_port *port, uint64_t publi
 
 int dp_port_meter_config(struct dp_port *port, uint64_t total_flow_rate_cap, uint64_t public_flow_rate_cap)
 {
-	if (dp_conf_get_nic_type() == DP_CONF_NIC_TYPE_TAP)
```
PlagueCZ, line 509: Maybe it would be better to switch the condition around to indicate that "this only works for 'hw'" option and is skipped for everything else.

src/dp_util.c
```diff
@@ -89,7 +89,7 @@ static int get_num_of_vfs_pattern(void)
 
 int dp_get_num_of_vfs(void)
 {
-	int vfs = dp_conf_get_nic_type() == DP_CONF_NIC_TYPE_TAP
```
PlagueCZ, line 92: And here it should then mean "sriov is only supported for 'hw' option"

include/dp_conf_opts.h
```diff
@@ -11,6 +11,7 @@
 enum dp_conf_nic_type {
 	DP_CONF_NIC_TYPE_HW,
 	DP_CONF_NIC_TYPE_TAP,
+	DP_CONF_NIC_TYPE_BLUEFIELD2,
```
PlagueCZ, line 14: Maybe it's not a good idea due to a breaking change, but HW/TAP/Bluefield seems strange, as previously this meant a switch between "real" and "virtual" cards, it now is more of a "type". Then I would suggest changing "hw" to "nvidia" or "mellanox" or "sriov".

The breaking change should only be visible in the docker container which is part of this repository, so it should not break much and as for OSC, there is the whole ironcore switch so breaks are expected anyway.

PlagueCZ (commented):
The change is perfectly functional, but maybe we can use this opportunity to make the option slightly better for the future?

include/dp_conf_opts.h
guvenc: Thanks for the suggestions. I think breaking change is not so problematic in this case.
I like Mellanox better. So we can simply change HW with Mellanox as below:

> enum dp_conf_nic_type {
	DP_CONF_NIC_TYPE_MELLANOX,
	DP_CONF_NIC_TYPE_TAP,
	DP_CONF_NIC_TYPE_BLUEFIELD2,
}

src/dp_util.c
guvenc: Yeah, I can turn the conditions around in both cases (also the one above), it would look more readable and nicer.

src/dp_port.c
PlagueCZ: Everything now looks good. But maybe we should emit a warning here so that there is a log line about the metering not being in effect due to the NIC type?

guvenc: Done, would make sense. Maybe better no warning when metering values are are both zero ?

PlagueCZ: Ah true, especially since calling this function with zeroes is used for rollback/disable of metering.

PlagueCZ (approved):

PR #486, opened 2024-01-30
Telemetry test should check for a pattern of string

guvenc:
On Bluefield2 cards, TAP devices do not start always with rx-0-0 so it is better to check for a string pattern and expected number of nodes instead of strictly checking for a specific string.


PlagueCZ (approved):
Makes sense as TAP was always from zero (since only test used those) and I had Mellanox on my desktop first, other setups were never tested, good catch.

PR #487, opened 2024-02-01
Support arm64 in Docker and github workflow

guvenc:
Support arm64 in Docker and github workflow

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,8 +76,8 @@ RUN cd $DPDK_DIR/build && ninja
 RUN cd $DPDK_DIR/build && ninja install
 
 # Get companion binaries from other repos
-ADD https://github.com/ironcore-dev/dpservice-cli/releases/download/v${CLI_VERSION}/github.com.ironcore-dev.dpservice-cli_${CLI_VERSION}_linux_amd64.tar.gz dpservice-cli.tgz
-ADD https://github.com/ironcore-dev/prometheus-dpdk-exporter/releases/download/v${EXPORTER_VERSION}/prometheus-dpdk-exporter_${EXPORTER_VERSION}_linux_amd64.tar.gz exporter.tgz
+ADD https://github.com/ironcore-dev/dpservice-cli/releases/download/v${CLI_VERSION}/github.com.ironcore-dev.dpservice-cli_${CLI_VERSION}_linux_${OSARCH}.tar.gz dpservice-cli.tgz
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,10 +1,12 @@
 # Build image with DPDK, etc.
-FROM debian:12-slim AS builder
+FROM --platform=${TARGETPLATFORM} debian:12-slim AS builder
```
PlagueCZ, line 2: This prints a warning for the default `docker build .` command. Not sure if this can be solved easily. The build itself seems to work though.

Dockerfile
```diff
@@ -41,8 +43,7 @@ pkg-config \
 protobuf-compiler-grpc \
 libgrpc++1.51 \
 libgrpc++-dev \
-libpcap0.8-dev \
-linux-headers-${OSARCH}
```
PlagueCZ, line 45: I could not test in live environment, so do not know what this will change. I guess nobody knows anymore why this was here?

PlagueCZ (commented):
What bout the other targets? I guess that bluefield does not use the "production" target with a pod image, so that is why it'S not changed here? What about the tester? I guess it's not needed at this stage to make the CI/CD take up twice the time for tests that can be run once?

Dockerfile
guvenc: Could you please use the docker ` buildx` to build ? This is the recommended way to build multi architecture docker images anyways. The command would be then `docker buildx build .`
Does buildx work fine ? buildx should automatically provide the builtin variables like "TARGETPLATFROM" or TARGETARCH or it can be also specified with "--platform linux/amd64" in local builds.

Dockerfile
guvenc: This was never included before, as the OSARCH was not set in old builds without the changes in this PR.
That is why I removed it.
If you take a look at the older builds without this change, you can see that it was never installed:
https://github.com/ironcore-dev/dpservice/actions/runs/7744421590/job/21118124520
image
image

guvenc:
@PlagueCZ Bluefield will use the production target. It is just the same docker image only difference is the ARM compilation.
I will omit the tester image for ARM, at least for the beginning. Only x86 image will be tested and the ARM image should at least compile.

Dockerfile
PlagueCZ: Cannot directly test on docker as we are using `podman` everywhere. In podman at least `buildx` is just an alias and the output is the same. But as image compilation work with just this warning (which in CI/CD will be solved), it's not a blocker.

guvenc: I would see this point as "acceptable" then as it doesnt break the podman workflow and we use buildx with github workflows so the CI/CD in github actions needs to cope with docker and buildx.

PlagueCZ (approved):

PR #488, opened 2024-02-02
Ignore sysfs  error for metering config due to unknow linux 6.x issue

byteocean:
To make sure interface can be created on linux 6.x machines, if corresponding sysfs file cannot be found (more precisely, the who sriov directory does not exist), the configuration for metering parameter of total traffic is ignored for now. The reason why sriov directory does not exist or other approach will be investigated further.

### DIFF
--- a/src/dp_port.c
+++ b/src/dp_port.c
@@ -506,6 +506,8 @@ static int dp_port_public_flow_meter_config(struct dp_port *port, uint64_t publi
 
 int dp_port_meter_config(struct dp_port *port, uint64_t total_flow_rate_cap, uint64_t public_flow_rate_cap)
 {
+	int err;
+
 	if (dp_conf_get_nic_type() != DP_CONF_NIC_TYPE_MELLANOX) {
 		if (public_flow_rate_cap != 0 || total_flow_rate_cap != 0)
 			DPS_LOG_WARNING("Metering config will not take effect due to the NIC type",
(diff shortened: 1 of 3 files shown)


src/dp_util.c
```diff
@@ -194,7 +194,7 @@ int dp_set_vf_rate_limit(uint16_t port_id, uint64_t rate)
 
 	if (res < 0 || (unsigned)res >= (int)sizeof(filename)) {
 		DPS_LOG_ERR("Failed to generate SR-IOV sysfs path or SR-IOV sysfs path to vf's max tx rate is too long");
-		return DP_ERROR;
+		return DP_GRPC_ERR_NOT_FOUND; // on Linux 6.x.y, this path/file does not seem to exist. thus for now, a not found error is returned and ignored by the caller
```
PlagueCZ, line 197: This should be -ENOENT to emulate what DPDK does for missing entries. GRPC errors are using special values tied to metalnet/golang.

Also this is not the right place, this is `snprintf` failure, the right place is few lines down where `fopen` fails, in which case you can actually just return `-errno` like DPDK does.

Though seeing the code itself, I don't think you need to even differentiate because unless `fopen` fails, the only error is calling this function on invalid port (will fail down the line anyway) or `snprintf` failing (i.e. path is too long) which is actually again just a sysfs path problem.

src/dp_port.c
```diff
@@ -519,11 +521,17 @@ int dp_port_meter_config(struct dp_port *port, uint64_t total_flow_rate_cap, uin
 		return DP_ERROR;
 	}
 
-	if (DP_FAILED(dp_port_total_flow_meter_config(port, total_flow_rate_cap))) {
-		DPS_LOG_ERR("Cannot set total flow meter", DP_LOG_PORT(port));
-		return DP_ERROR;
+	err = dp_port_total_flow_meter_config(port, total_flow_rate_cap);
+	if (DP_FAILED(err)) {
+		if (err == DP_GRPC_ERR_NOT_FOUND)
+			DPS_LOG_WARNING("Cannot find sysfs path or file to regulate traffic rate, thus total flow rate metering is ignored", DP_LOG_PORT(port));
```
PlagueCZ, line 527: Unless specifically asking for `-ENOENT` This error should not mention the reason of failure, just that metering is ignored. Because currently (assuming we check `fopen`) this can also fail due to file access permissions, etc.

PlagueCZ (changes requested):

PlagueCZ:
I think that Bluefield2 (shown on networking weekly call) is running Linux 5.X and it also is missing the sysfs path, so maybe this is not a Linux 6.X problem at all?

byteocean:
@PlagueCZ yeah, it could be. but the change here does not really detect the version of linux kernel, but instead, if it cannot find this sysfs file, simply ignore this part of hardware configuration and continue with other configs.

src/dp_port.c
byteocean: you are right regarding other failure due to access permissions. and I would like to explicitly show that part of metering configuration cannot be done until we a solution is found. thus, I used `access` to explicitly to detect the existence of this sysfile, and return an -ENOENT when needed.

src/dp_util.c
byteocean: fixed.

PlagueCZ (approved):
As a temporary solution this seems fine. In the log run maybe differentiating between EACCES and ENOENT would be nice.

guvenc (approved):

PR #491, opened 2024-02-06
Fix failed continuing echo to IPv6 LB

byteocean:
only the first ping to IPv6 LB works. fix with condition change in the code.


guvenc (changes requested):
Good catch. Looks good. When we enhance the test for ipv6, would be good to enhance it also for ipv4 in a symmetric way.

guvenc (approved):

PR #494, opened 2024-02-08
Adjust prepare.sh script to support Bluefield 2

guvenc:
Adjust prepare.sh in such a way that it supports bluefield 2 card in a backward compatible way and still works with mellanox cards.
Bluefield cards are configured on the host side but the Bluefield card side still needs the dp_service.conf so the script should 
generate dp_service.conf without complaining about the missing sysfs file parts on the bluefield.

### DIFF
--- a/hack/prepare.sh
+++ b/hack/prepare.sh
@@ -9,8 +9,10 @@ set -Eeuo pipefail
 # Use as /usr/local/sbin/dp-prepare.sh
 #
 
+BLUEFIELD_IDENTIFIERS=("MT_0000000543")
 NUMVFS=126
 CONFIG="/tmp/dp_service.conf"
+IS_BLUEFIELD=false
 
 function log() {
 	echo "$(date +"%Y-%m-%d_%H-%M-%S-%3N") $1"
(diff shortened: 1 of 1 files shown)


hack/prepare.sh
```diff
@@ -25,6 +26,17 @@ function get_pf() {
 	readarray -t devs < <(devlink dev | awk -F/ '{print $2}')
 }
 
+function detect_card_type() {
+	local pf="${devs[0]}"
+	if devlink dev info pci/$pf | grep -q "fw.psid MT_0000000543"; then
```
byteocean, line 31: The changes look good! Just wondering if this PSID field needs to be set as configurable or at least to use a variable. As I can tell from this [website](https://docs.nvidia.com/networking/display/bluefield2firmwarev24401000/firmware+compatible+products), different serials of BlueField-2 cards (maybe with different prices), PSID could be slightly different.

guvenc: Valid point. I introduced an "array" of Bluefield identifiers now, so that "array" can be extended with the Bluefield identifiers that we want to support easily and the detection mechanism can stay the same.

byteocean: cool

byteocean (approved):

PR #504, opened 2024-03-01
Identify whether the prepare.sh script on host or on bluefield runs

guvenc:
The Bluefield card is visible to the host x86 system where it is connected and the prepare script also runs on the host system. (not the dpservice but just the prepare script)

Therefore the script needs to be able to differentiate between the cases running on the bluefield directly and on the x86 host where the bluefield is connected and there is also the original case x86 host with a Mellanox card. They all need slightly different treatments in the script.


PlagueCZ (approved):
Getting more complex with each iteration but cannot be helped. Change looks good

PR #506, opened 2024-03-05
Migrate to DPDK 23.11

byteocean:
Migrate to DPDK 23.11


byteocean:
Didn't really capture an obvious issue in the lab environment for both non-offloading and offloading. thus it is made as ready for review.

src/grpc/dp_grpc_thread.cpp
```diff
@@ -29,7 +29,8 @@ static void *dp_grpc_main_loop(__rte_unused void *arg)
 		rte_exit(EXIT_FAILURE, "Cannot run without working gRPC server\n");
 
 	GRPCService::Cleanup();
-	return NULL;
+
```
PlagueCZ, line 32: This should either be done for all functions here or none of them. It's meant to be a single "block" of state-change and return OK, like the other functions.

src/dp_service.c
```diff
@@ -187,11 +187,11 @@ static void free_interfaces(void)
 	dp_ifaces_free();
 	dp_flow_free();
 	dp_telemetry_free();
+	dp_ports_free();
```
PlagueCZ, line 190: Since this was a single commit, it must be helping, but I cannot see why that is. Do you have any details as to why this helps DPDK to clean up properly?

src/dp_port.c
```diff
@@ -100,14 +100,18 @@ static int dp_port_init_ethdev(struct dp_port *port, struct rte_eth_dev_info *de
...
+	else
+		nr_hairpin_queues = port->is_pf
+			? (uint16_t)(DP_NR_PF_HAIRPIN_RX_TX_QUEUES + DP_NR_VF_HAIRPIN_RX_TX_QUEUES * dp_layer->num_of_vfs)
+			: DP_NR_VF_HAIRPIN_RX_TX_QUEUES;
 
 	ret = rte_eth_dev_configure(port->port_id,
 								DP_NR_STD_RX_QUEUES + nr_hairpin_queues,
 								DP_NR_STD_TX_QUEUES + nr_hairpin_queues,
 								&port_conf);
+
```
PlagueCZ, line 114: We usually do 
```
ret = xxx();
if DP_FAILED(ret)
   return ret;
```
so this empty line actually breaks the usual pattern.

PlagueCZ (commented):
I tried really hard, but only found a few empty lines to comment on, looks great. I should also try updating my DPDK and compiling tomorrow.

src/grpc/dp_grpc_thread.cpp
byteocean: removed.

src/dp_port.c
byteocean: removed.

src/dp_service.c
byteocean: In `dp_ports_free()`, the ctx of a graph node is accessed to set a value. this ctx is actually freed together with graph node. Due to the code change within `rte_graph_destroy()`, this invalid access gets exposed.

PlagueCZ (approved):
Nice find with the now not needed context table

guvenc (approved):
Looks good. Thanks.

PR #509, opened 2024-03-27
Introduce maglev hashing for loadbalancer backend selection

guvenc:
Implement maglev hashing for loadbalancer backend selection. Removed round robin completely.
Moved one complex loadbalancer test to "flow" suite, as the flow timeouts need to be short 
for maglev testing scenarios.


test/xtratest_flow_timeout.py
```diff
@@ -132,3 +132,59 @@ def test_external_lb_relay_timeout(prepare_ipv4, grpc_client, fast_flow_timeout)
...
+	grpc_client.dellbprefix(VM2.name, lb_pfx)
+	grpc_client.dellbtarget(lb_name, lb_vm1_ul_ipv6)
+	grpc_client.dellbprefix(VM1.name, lb_pfx)
+	grpc_client.dellb(lb_name)
+
+	grpc_client.delfwallrule(VM2.name, "fw0-vm2")
+	grpc_client.delfwallrule(VM1.name, "fw0-vm1")
+
+	# NOTE: this test, just like in test_pf_to_vf.py
+	# cannot be run twice in a row, since the flows need to age-out
```
PlagueCZ, line 190: I get why you needed to do the maglev test in this test file, but I think the original LB-VIP test should still be present in the standard test file too, as the `xtratest_flow_timeout.py` should only test the special cases that use timeout.

I would suggest removing the NAT part here and using the old test code without roundroubin/malgev part in the original location.

test/helpers.py
```diff
@@ -165,3 +165,44 @@ def age_out_flows():
...
+	srv_reply = (Ether(dst=srv_pkt[Ether].src, src=srv_pkt[Ether].dst, type=0x0800) /
+				 IP(dst=srv_pkt[IP].src, src=srv_pkt[IP].dst) /
+				 TCP(sport=srv_pkt[TCP].dport, dport=srv_pkt[TCP].sport))
+	delayed_sendp(srv_reply, vf_tap)
+	# HTTP response from the router on vm(VIP)
+	vm_reply = sniff_packet(vm.tap, is_tcp_pkt)
+	assert vm_reply[IP].dst == vm.ip, \
+		f"Invalid VIPped destination IP {vm_reply[IP].dst}"
+	assert vm_reply[TCP].sport == 80, \
+		f"Invalid server reply port {vm_reply[TCP].sport}"
```
PlagueCZ, line 208: Good idea to separate the LB test code if it's used in multiple places, but please put into something akin to `tcp_tester.py` as the helpers should really be for cases used globally in the whole test suite. At least that was the idea.

src/dp_maglev.c
```diff
@@ -0,0 +1,210 @@
...
+			break;
+		}
+	}
+
+	if (lbval->back_end_cnt == 0)
+		return DP_OK;
+	else
+		return dp_maglev_calc_hash_lookup_table(lbval);
+}
+
```
PlagueCZ, line 210: Hopefully I'm right and this means no offense, but this code is surely based on a paper or some other research telling you how best to implement this algorithm. Which is perfectly fine, just that it is hard to understand due to lack of deeper understanding on the reader's part. 
I do not think we should pollute the code with comments, but a link or a reference of the implementation paper at the top would be a great addition.

include/dp_maglev.h
```diff
@@ -0,0 +1,20 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+#include <stdint.h>
+#include "dp_util.h"
+#include "dp_lb.h"
+
+int dp_delete_maglev_backend(struct lb_value *lbval, const uint8_t *delete_server);
+int dp_add_maglev_backend(struct lb_value *lbval, const uint8_t *new_server);
```
PlagueCZ, line 15: I'm a bit confused as to what `server` means in this context, it seems to suggest to me that this is somehow distributed. Or it's just the source material nomenclature leftover. We use `back_ip` everywhere in place of this, unless I am misunderstanding how this works.

Dockerfile
```diff
@@ -101,7 +101,7 @@ FROM builder AS testbuilder
 ARG DPSERVICE_FEATURES=""
 RUN meson setup release_build $DPSERVICE_FEATURES --buildtype=release && ninja -C release_build
 RUN CC=clang CXX=clang++ meson setup clang_build $DPSERVICE_FEATURES && ninja -C clang_build
-RUN meson setup xtratest_build $DPSERVICE_FEATURES -Denable_tests=true && ninja -C xtratest_build
```
PlagueCZ, line 104: Is this still needed or just debugging leftover? The grpc wrapper python script should return all underlay addresses so that they can be saved for later and used, no need o have them static for the test suite usually.

src/dp_maglev.c
```diff
@@ -0,0 +1,210 @@
...
+	return hash;
+}
+
+static int *dp_maglev_permutation(struct lb_value *lbval)
+{
+	int  i, j, size, *permutation;
+	uint32_t offset, skip;
+
+	size = lbval->back_end_cnt * DP_LB_MAGLEV_LOOKUP_SIZE;
+	permutation = rte_zmalloc("maglev_perm", sizeof(int) * size, RTE_CACHE_LINE_SIZE);
```
PlagueCZ, line 81: Both here and the other allocation below do not log anything, though improbable, it would still be helpful to have in the log.

src/dp_maglev.c
```diff
@@ -0,0 +1,210 @@
...
+				if (lbval->maglev_hash[pos] == DP_MAGLEV_POS_FREE) {
+					lbval->maglev_hash[pos] = (int16_t)i;
+					num++;
+					if (num == DP_LB_MAGLEV_LOOKUP_SIZE)
+						goto out;
+					break;
+				}
+			}
+		}
+	}
```
PlagueCZ, line 121: I guess this is Google's algorithm, but still, an infinite loop just raises alarms for me.

Maybe having an always inline `dp_maglev_populate_iteration()` and then `while (dp_maglev_populate_iteration())` which also eliminates the `goto` would make this seem a bit more tame?

But if this is more of a 1:1 code from Google and there is a reference to such a document, then it would be best left as-is to resemble the original.

PlagueCZ (commented):
In spite of the number of comments, this PR looks really nice! Good separation into a submodule and except for the complex algorithm itself it's really clean and simple.

byteocean (commented):
I was able to closely look at this maglev hashing part, especially the algorithm parts, which are in general nicely implemented. there are two thoughts that I can share to improve the overall robustness of algorithm implementation.

include/dp_error.h
```diff
@@ -70,6 +70,8 @@ const char *dp_strerror_verbose(int error);
 	ERR(VNF_DELETE,							403) \
 	ERR(NO_BACKIP,							421) \
 	ERR(NO_LB,								422) \
+	ERR(BACKIP_ADD_FAILED,					423) \
+	ERR(BACKIP_DEL_FAILED,					424) \
```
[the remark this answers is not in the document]
guvenc: Done

src/dp_lb.c
```diff
@@ -266,17 +211,21 @@ uint8_t *dp_lb_get_backend_ip(struct flow_key *flow_key, uint32_t vni)
...
-	lb_val->last_sel_pos = (uint16_t)pos;
+	for (k = 0; k < DP_LB_MAX_PORTS; k++) {
+		if ((lb_val->ports[k].port == htons(flow_key->port_dst)) && (lb_val->ports[k].protocol == flow_key->proto))
+			break;
+		if (lb_val->ports[k].port == 0)
+			return NULL;
+	}
+	/* Port table full and we didnt get a port match */
+	if (k == DP_LB_MAX_PORTS)
+		return NULL;
```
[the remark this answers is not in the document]
guvenc: Done

test/xtratest_flow_timeout.py
guvenc: We can not run the original LB-VIP test as with the new maglev implementation and random generated underlay ipv6 addresses, the backend selection is not deterministic.
I am running this here as this one is built with disabled random underlay ipv6 generation.

test/helpers.py
guvenc: Done

src/dp_maglev.c
guvenc: Put a comment at the beginning of the file that this code is based on maglev paper. Googling it should not be a problem, if someone wants to know more.

include/dp_maglev.h
guvenc: Changed the names to back_ip to be consistent with the rest of the code.

Dockerfile
guvenc: maglev hashing doesnt behave deterministic with shuffled underlay ipv6 addresses therefore it is still needed.

src/dp_maglev.c
guvenc: Done

include/dp_lb.h
```diff
@@ -11,13 +11,10 @@ extern "C" {
...
-#define DP_LB_MAX_IPS_PER_VIP	64
-
-#define DP_LB_OFF	0
-#define DP_LB_ON	1
-#define DP_LB_LB	2
-#define DP_LB_DLB	3
+#define DP_LB_TABLE_MAX				256
+#define DP_LB_MAX_IPS_PER_VIP		64
+/* Needs to be a prime number at least 2xDP_LB_MAX_IPS_PER_VIP for an uniform distrubition */
+#define DP_LB_MAGLEV_LOOKUP_SIZE	131
```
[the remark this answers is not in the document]
guvenc: Done

src/dp_maglev.c
```diff
@@ -0,0 +1,210 @@
...
+		offset = dp_murmur_hash2(lbval->back_end_ips[i]) % DP_LB_MAGLEV_LOOKUP_SIZE;
+		skip = (dp_djb_hash(lbval->back_end_ips[i]) % (DP_LB_MAGLEV_LOOKUP_SIZE - 1)) + 1;
+
+		for (j = 0; j < DP_LB_MAGLEV_LOOKUP_SIZE; j++)
+			permutation[i * DP_LB_MAGLEV_LOOKUP_SIZE + j] = (offset + j * skip) % DP_LB_MAGLEV_LOOKUP_SIZE;
+	}
+	return permutation;
+}
+
+static int dp_maglev_populate(struct lb_value *lbval, int *permutation)
```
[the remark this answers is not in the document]
guvenc: Done

test/xtratest_flow_timeout.py
PlagueCZ: That's what I am saying, that the maglev test needs to stay as-is here. The original location should only test whether communication NAT->LB in differing VNIs works.

Meaning the original test would not contain any add/remove LB targets, no flow aging, and you would only use one target, thus no randomness is present.

PlagueCZ (approved):
Looks great

byteocean (approved):

Issue #513, opened 2024-04-05
GitHub test action sometimes fails on VNF entry deletion


PlagueCZ:
On rare instances (currently it seems that something in that the time window must align for this to be at least sometimes reproducible by re-running the `test` action), the tester docker image fails on prefix or lbprefix deletion.

Posting the lbprefix deletion because its code does less actions than the alias prefix counterpart, yet the error is the same:
```
----------------------------- Captured stdout call -----------------------------
dpservice-cli add lbprefix --interface-id=vm2 --prefix=2a10:defe:e01f:f4::2/128
2024-04-04 13:27:15.318 2(grpc) I GRPC: Adding loadbalancer target prefix, interface_id: vm2, prefix: 2a10:defe:e01f:f4::2, prefix_len: 128 [../src/grpc/dp_async_grpc.cpp:857:FillRequest()]
 > {"kind":"LoadBalancerPrefix","metadata":{"interface_id":"vm2"},"spec":{"prefix":"2a10:defe:e01f:f4::2/128","underlay_route":"fc00:1::7f:0:fb"},"status":{"code":0,"message":"Success"}}
dpservice-cli list lbprefixes --interface-id=vm2
2024-04-04 13:27:15.326 2(grpc) I GRPC: Listing loadbalancer target prefixes, interface_id: vm2 [../src/grpc/dp_async_grpc.cpp:899:FillRequest()]
 > {"kind":"LoadBalancerPrefixList","metadata":{"interface_id":"vm2"},"status":{"code":0,"message":"Success"},"items":[{"kind":"LoadBalancerPrefix","metadata":{"interface_id":"vm2"},"spec":{"prefix":"2a10:defe:e01f:f4::2/128","underlay_route":"fc00:1::7f:0:fb"},"status":{"code":0,"message":""}}]}
dpservice-cli add lbprefix --interface-id=invalid_name --prefix=2a10:defe:e01f:f4::2/128
2024-04-04 13:27:15.338 2(grpc) I GRPC: Adding loadbalancer target prefix, interface_id: invalid_name, prefix: 2a10:defe:e01f:f4::2, prefix_len: 128 [../src/grpc/dp_async_grpc.cpp:857:FillRequest()]
2024-04-04 13:27:15.338 3(worker) W GRPC: Failed request, grpc_request: 30, grpc_error: 205, grpc_message: NO_VM [../src/grpc/dp_grpc_impl.c:1146:dp_process_request()]
[...]

PR #514, opened 2024-04-05
Feature/vnf c small improvements

PlagueCZ:
While trying to find any possible bug in VNF code due to #513 I made the following changes. Thee only functional change is the use of hashing before duplicate calls to hash tables, but given the fact that this code is not hot path, it's not of high importance. 

Most changes are thus to streamline the code and remove some small inconsistencies.


guvenc (approved):
Thanks. Looks good to me.

PR #515, opened 2024-04-05
Test PR for pre-dpdk23 test

PlagueCZ:
The idea is to find #513 happening before update to DPDK 23 so we can rule that out as a problem.


PlagueCZ:
Needed for #513, which is closed now.

PR #519, opened 2024-04-09
Fix packed structure being used for nat portmap hash key

### LINKED ISSUE #518: `struct netnat_portmap_key` is not properly packed
Going over the newly introduced `struct dp_ip_address` that holds both IPv4 and IPv6, I noticed it is included in `struct netnat_portmap_key` which is used as a hashtable key.

Even though the key is packed, the packing does not propagate, thus the key is actually not properly packed and thus it's hash can be undefined.

Using this code in `dp_nat.c:608:dp_allocate_network_snat_port()`:
```
printf("key size: %lu\n", sizeof(portmap_key));
printf(" ip size: %lu\n", sizeof(portmap_key.src_ip));
printf(" tp size: %lu\n", sizeof(portmap_key.src_ip.ip_type));
assert(false);
```
The output is:
```
key size: 26
 ip size: 20
 tp size: 2
dpservice-bin: ../src/dp_nat.c:624: dp_allocate_network_snat_port: Assertion `false' failed.
```
Given that `struct dp_ip_address` is 16B of address and 2B of type, its size is clearly padded.

PlagueCZ:
I simply unrolled the `dp_ip_address` structure into the key, like `struct lb_key` does. There will be a bigger upcoming change to the `dp_ip_address` itself that will touch this code again, so I did not optimize anything.

I also noticed that the IP type was never set for this key, which would only be a problem when IPv4 followed by all zeroes would match an IPv6 I guess, but I still added the type.

I reordered the key for better alignment, that's why other fields are moved too.

The change in test is caused by a different hash for the same data (since I reordered and removed padding), thus a resulting NAT port is chosen differently.

Fixes #518

### DIFF
--- a/include/dp_nat.h
+++ b/include/dp_nat.h
@@ -50,9 +50,13 @@ struct dnat_data {
 };
 
 struct netnat_portmap_key {
-	struct dp_ip_address	src_ip;
-	uint32_t				vni;
-	uint16_t				iface_src_port;
+	union {
+		uint8_t		src_ipv6[DP_IPV6_ADDR_SIZE];
+		uint32_t	src_ipv4;
+	};
+	uint16_t		src_type;
+	uint16_t		iface_src_port;
+	uint32_t		vni;
 } __rte_packed;
 
 struct netnat_portmap_data {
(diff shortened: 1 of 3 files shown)


guvenc (approved):
Good catch. Looks good.

PR #521, opened 2024-04-10
Discussion about potential problems in IPv4/IPv6 structures

PlagueCZ:
I noticed the following inconsistencies when trying to add some safer wrappers around all the places where (now dual IPv4/IPv6) IP address needs to be present in some hash key.

The last place that needs it is `struct dp_flow_key` and while doing the change I noticed that sometimes the code is not updated to the new fact that the IP can be both IPv4 and IPv6.

### DIFF
--- a/include/dp_flow.h
+++ b/include/dp_flow.h
@@ -69,6 +69,9 @@ enum dp_flow_tcp_state {
 };
 
 struct flow_key {
+	// my proposed refactor would be much simpler if there could be l3_type for *both* l3_src and l3_dst
+	// (as part of a common packed struct dp_ip_addr_key)
+	// I realize that having DST and SRC types differ makes no sense, so it would waste one byte, but all handling functions would be cleaner
 	union {
 		uint32_t	ip4;
 		uint8_t		ip6[DP_IPV6_ADDR_SIZE];
--- a/src/dp_cntrack.c
+++ b/src/dp_cntrack.c
@@ -162,6 +162,7 @@ static __rte_always_inline struct flow_value *flow_table_insert_entry(struct flo
 	flow_val->timeout_value = flow_timeout;
 	flow_val->created_port_id = port->port_id;
 
+	// what if key->l3_type == RTE_ETHER_TYPE_IPV6?
 	pfx_ip.ip_type = RTE_ETHER_TYPE_IPV4;
 	pfx_ip.ipv4 = key->l3_dst.ip4;
 	/* Target ip of the traffic is an alias prefix of a VM in the same VNI on this dp-service */
--- a/src/dp_flow.c
+++ b/src/dp_flow.c
@@ -100,6 +100,7 @@ static __rte_always_inline int dp_build_icmp_flow_key(const struct dp_flow *df, 
 			return DP_ERROR;
 		}
 
+		// changing l3_dst and l3_src without setting l3_type to IPV4
 		key->l3_dst.ip4 = ntohl(icmp_err_ip_info.err_ipv4_hdr->src_addr);
 		key->l3_src.ip4 = ntohl(icmp_err_ip_info.err_ipv4_hdr->dst_addr);
 
@@ -123,6 +124,7 @@ static __rte_always_inline void dp_mark_vnf_type(struct dp_flow *df, const struc
 	struct snat_data *s_data;
 	struct dp_ip_address pfx_ip;
 
+	// what if key->l3_type == RTE_ETHER_TYPE_IPV6? (see caller)
 	memset(&pfx_ip, 0, sizeof(pfx_ip));
 	pfx_ip.ip_type = RTE_ETHER_TYPE_IPV4;
 	pfx_ip.ipv4 = key->l3_src.ip4;
@@ -133,6 +135,7 @@ static __rte_always_inline void dp_mark_vnf_type(struct dp_flow *df, const struc
 		else
 			key->vnf_type = DP_VNF_TYPE_UNDEFINED;
 	} else if (key->l3_type == RTE_ETHER_TYPE_IPV6) {
+		// does this need both l3_src and l3_dst to be IPV6 or only l3_dst (important for a proposed solution to other problems)
 		if (dp_is_ip6_in_nat64_range(key->l3_dst.ip6))
 			key->vnf_type = DP_VNF_TYPE_NAT;
 	} else {
@@ -166,6 +169,8 @@ int dp_build_flow_key(struct flow_key *key /* out */, struct rte_mbuf *m /* in *
 		rte_memcpy(key->l3_src.ip6, df->src.src_addr6, sizeof(key->l3_src.ip6));
 		break;
 	default:
+		// what if key->l3_type == RTE_ETHER_TYPE_IPV6?
+		// also, whouldn't this just return DP_ERROR?
 		key->l3_dst.ip4 = 0;
 		key->l3_src.ip4 = 0;
 		break;
@@ -199,6 +204,7 @@ int dp_build_flow_key(struct flow_key *key /* out */, struct rte_mbuf *m /* in *
 	default:
 		key->port_dst = 0;
 		key->src.port_src = 0;
+		// should this not return DP_ERROR?
 		break;
 	}
 
@@ -207,6 +213,7 @@ int dp_build_flow_key(struct flow_key *key /* out */, struct rte_mbuf *m /* in *
 
 void dp_invert_flow_key(const struct flow_key *key /* in */, uint16_t l3_type /* in */, struct flow_key *inv_key /* out */)
 {
+	// since we are inverting the key, couldn't this ask key->l3_type instead? (will make more stuff easier)
 	if (l3_type == RTE_ETHER_TYPE_IPV4) {
 		inv_key->l3_src.ip4 = key->l3_dst.ip4;
 		inv_key->l3_dst.ip4 = key->l3_src.ip4;
@@ -294,6 +301,7 @@ int dp_get_flow(const struct flow_key *key, struct flow_value **p_flow_val)
 bool dp_are_flows_identical(const struct flow_key *key1, const struct flow_key *key2)
 {
 	return key1->proto == key2->proto
+		// this should compare ipv6 too
 		&& key1->l3_src.ip4 == key2->l3_src.ip4
 		&& key1->l3_dst.ip4 == key2->l3_dst.ip4
 		&& key1->port_dst == key2->port_dst
@@ -511,6 +519,7 @@ void dp_remove_neighnat_flows(uint32_t ipv4, uint32_t vni, uint16_t min_port, ui
 			DPS_LOG_ERR("Iterating flow table failed while removing NAT flows", DP_LOG_RET(ret));
 			return;
 		}
+		// what if key->l3_type == RTE_ETHER_TYPE_IPV6?
 		if (next_key->vni == vni && next_key->l3_dst.ip4 == ipv4
 			&& next_key->port_dst >= min_port && next_key->port_dst < max_port
 		) {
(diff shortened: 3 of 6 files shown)


include/dp_flow.h
```diff
@@ -69,6 +69,9 @@ enum dp_flow_tcp_state {
 };
 
 struct flow_key {
+	// my proposed refactor would be much simpler if there could be l3_type for *both* l3_src and l3_dst
+	// (as part of a common packed struct dp_ip_addr_key)
+	// I realize that having DST and SRC types differ makes no sense, so it would waste one byte, but all handling functions would be cleaner
```
PlagueCZ, line 74: This is just a question for the refactor I have prepared. Since more places use the IPv4/IPv6 in a hash key, I would have a common structure (packed) to use along with handler functions that take care of the whole "make sure you zero the rest and set type properly". 

No I would like to also use these safer wrappers for `flow_key` but it only has one type and two IPs, thus I would propose to sacrifice one byte to simply have the two common structures used.

src/dp_cntrack.c
```diff
@@ -162,6 +162,7 @@ static __rte_always_inline struct flow_value *flow_table_insert_entry(struct flo
 	flow_val->timeout_value = flow_timeout;
 	flow_val->created_port_id = port->port_id;
 
+	// what if key->l3_type == RTE_ETHER_TYPE_IPV6?
 	pfx_ip.ip_type = RTE_ETHER_TYPE_IPV4;
 	pfx_ip.ipv4 = key->l3_dst.ip4;
```
PlagueCZ, line 167: As this is only used for `dp_vnf_lbprefix_exists()` I guess just add the test for `l3_type` in the condition?

src/dp_flow.c
```diff
@@ -100,6 +100,7 @@ static __rte_always_inline int dp_build_icmp_flow_key(const struct dp_flow *df,
 			return DP_ERROR;
 		}
 
+		// changing l3_dst and l3_src without setting l3_type to IPV4
 		key->l3_dst.ip4 = ntohl(icmp_err_ip_info.err_ipv4_hdr->src_addr);
 		key->l3_src.ip4 = ntohl(icmp_err_ip_info.err_ipv4_hdr->dst_addr);
```
PlagueCZ, line 105: Now this one can be "solved" by a comment if this code only fires for IPv4

src/dp_flow.c
```diff
@@ -123,6 +124,7 @@ static __rte_always_inline void dp_mark_vnf_type(struct dp_flow *df, const struc
 	struct snat_data *s_data;
 	struct dp_ip_address pfx_ip;
 
+	// what if key->l3_type == RTE_ETHER_TYPE_IPV6? (see caller)
 	memset(&pfx_ip, 0, sizeof(pfx_ip));
 	pfx_ip.ip_type = RTE_ETHER_TYPE_IPV4;
 	pfx_ip.ipv4 = key->l3_src.ip4;
```
PlagueCZ, line 130: This is again just for LB prefix lookup, so maybe just change the condition below to check `l3_type` and only then pull the IPv4 out?

src/dp_flow.c
```diff
@@ -133,6 +135,7 @@ static __rte_always_inline void dp_mark_vnf_type(struct dp_flow *df, const struc
 		else
 			key->vnf_type = DP_VNF_TYPE_UNDEFINED;
 	} else if (key->l3_type == RTE_ETHER_TYPE_IPV6) {
+		// does this need both l3_src and l3_dst to be IPV6 or only l3_dst (important for a proposed solution to other problems)
 		if (dp_is_ip6_in_nat64_range(key->l3_dst.ip6))
 			key->vnf_type = DP_VNF_TYPE_NAT;
```
PlagueCZ, line 140: This is kind of bad wording, since of course both src and dest are of the same type here. But this is mostly a callback to the proposed change to `struct flow_key`

src/dp_flow.c
```diff
@@ -199,6 +204,7 @@ int dp_build_flow_key(struct flow_key *key /* out */, struct rte_mbuf *m /* in *
 	default:
 		key->port_dst = 0;
 		key->src.port_src = 0;
+		// should this not return DP_ERROR?
 		break;
 	}
```
PlagueCZ, line 209: These two I actually just thought I would return and see what happens, as I do not really know the implications right now

src/dp_flow.c
```diff
@@ -207,6 +213,7 @@ int dp_build_flow_key(struct flow_key *key /* out */, struct rte_mbuf *m /* in *
 
 void dp_invert_flow_key(const struct flow_key *key /* in */, uint16_t l3_type /* in */, struct flow_key *inv_key /* out */)
 {
+	// since we are inverting the key, couldn't this ask key->l3_type instead? (will make more stuff easier)
 	if (l3_type == RTE_ETHER_TYPE_IPV4) {
 		inv_key->l3_src.ip4 = key->l3_dst.ip4;
 		inv_key->l3_dst.ip4 = key->l3_src.ip4;
```
PlagueCZ, line 219: I think this is just a leftover from the times when `flow_key` had no `l3_type` and I can easily just drop the old argument

src/dp_flow.c
```diff
@@ -294,6 +301,7 @@ int dp_get_flow(const struct flow_key *key, struct flow_value **p_flow_val)
 bool dp_are_flows_identical(const struct flow_key *key1, const struct flow_key *key2)
 {
 	return key1->proto == key2->proto
+		// this should compare ipv6 too
 		&& key1->l3_src.ip4 == key2->l3_src.ip4
 		&& key1->l3_dst.ip4 == key2->l3_dst.ip4
 		&& key1->port_dst == key2->port_dst
```
PlagueCZ, line 307: Here I actually noticed that we are manually comparing every field of the key anyway. So I propose to move this into a `static __rte_always_inline` and compare it there in a fast way (using `uint64_t` comparisons). Because that is the fastest way to compare fixed-size blocks (for variable-size `rte_memcmp` is best).

src/dp_flow.c
```diff
@@ -531,6 +540,7 @@ void dp_remove_iface_flows(uint16_t port_id, uint32_t ipv4, uint32_t vni)
 			DPS_LOG_ERR("Iterating flow table failed while removing VM flows", DP_LOG_RET(ret));
 			return;
 		}
+		// what if key->l3_type == RTE_ETHER_TYPE_IPV6?
 		if (flow_val->created_port_id == port_id
 			|| (next_key->vni == vni && flow_val->flow_key[DP_FLOW_DIR_ORG].l3_dst.ip4 == ipv4)
 		) {
```
PlagueCZ, line 546: Both above can simply be fixed by adding the condition `l3_type == IPV4` since the code is looking for them specifically.

src/dp_nat.c
```diff
@@ -721,6 +721,7 @@ int dp_remove_network_snat_port(const struct flow_value *cntrack)
 	struct dp_port *created_port;
 	int ret;
 
+	// what if cntrack->flow_key[DP_FLOW_DIR_REPLY].l3_type == RTE_ETHER_TYPE_IPV6?
 	portoverload_tbl_key.nat_ip = cntrack->flow_key[DP_FLOW_DIR_REPLY].l3_dst.ip4;
 	portoverload_tbl_key.dst_ip = cntrack->flow_key[DP_FLOW_DIR_ORG].l3_dst.ip4;
 	portoverload_tbl_key.nat_port = cntrack->flow_key[DP_FLOW_DIR_REPLY].port_dst;
```
PlagueCZ, line 727: This one, an all following ones in `dnat_node` and `snat_node` are NAT related and I think they simply work because NAT only uses IPv4?

But unless you are familiar with the code base this will never be clear and will always seem problematic, so maybe just return error when the type is wrong? In this file it's no problem then the other ones are a hot path...

include/dp_flow.h
guvenc: Yes, it is ok to sacrifice 1 byte for ease of implementation and consistency.

src/dp_cntrack.c
guvenc: Adding the l3_type for comparison is fine.

src/dp_flow.c
guvenc: Yes, fired only in ICMP path which is ipv4 only.

src/dp_flow.c
guvenc: Yes, the suggestion is fine.

src/dp_flow.c
guvenc: It needs both l3_src and and l3_dst to be type IPv6 and it needs dest ipv6 to be in a specific range.

src/dp_flow.c
guvenc: What do you mean if key->l3_type == IPV6 ?
There is such a line before the switch `key->l3_type = df->l3_type;` and in this default branch of switch, it is impossible that it can be IPv6 as it would be another branch of the switch.
Returning error should be fine as it would  drop the packet and it will probably make us more robust against fuzzed packets during parsing.

src/dp_flow.c
guvenc: Yes, we can drop the old argument.

src/dp_flow.c
guvenc: Good idea.

src/dp_flow.c
guvenc: Yes

src/dp_nat.c
guvenc: Exactly. NAT only uses IPv4.
IPv6 is NATless apart from the fact that NAT64 but it works totally different then the ipv6 NAT.

Returning error for IPv6 should be fine. You can still check the call chain to have a rough idea what might be happening if you return error.

guvenc (commented):
Thanks for the refactoring.

src/dp_nat.c
PlagueCZ: There is a problem in NAT64 implementation here. in the NAT64 test `test_vf_to_pf.py::test_vf_to_pf_network_nat_icmpv6` (you can only run this one to see), the flow is not deleted here, because `portoverload_tbl_key.dst_ip = cntrack->flow_key[DP_FLOW_DIR_ORG].l3_dst.ip4;` is not actually IPv4, it's the NAT64 address.

If I copy the code from `dp_allocate_network_snat_port()` where you pull the IPv4 out from NAT64, the address is converted right, but unfortnuately there must be som followup problem because `portoverload_tbl_key.dst_port = flow_key_org->port_dst;` is ZERO for some reason, thus the flow is never deleted.

There is no error or check for it in the test suite so tests behave normally, but you can easily add debug outputs to see that a NAT64 flow is added in `dp_allocate_network_snat_port()` but never deleted in `dp_remove_network_snat_port` because of this.

guvenc: I need to revise/correct part of my previous statement. That is what happens, if you look at the code part, after a long while.
NAT is actually used by ipv6 in a similar way as in ipv4 for the NAT ipv4 address used in NAT64. Probably you figured that out, already, because of the code manipulating source ports.
To be honest, I can not tell you at the moment how to deal with the bug here but if it hinders your refactoring, can you leave this part as it is and we can fix it in a follow-up ? I will try to give it a more thought tomorrow.

PlagueCZ: It's not problem for the refactoring (tests still work), I can then create a separate issue on the ne code.

guvenc: Ok. This seems to be broken already in ipv4 and ipv6 introduction doesnt seem to make it "worse". (ipv6 seems to be putting the wrong dst ip address in the key in addition to what was already broken)
The same problem can be observed with the ipv4 test `test_vf_to_pf_network_nat_icmp`
I will open a bug to track it and I think @byteocean can shed a light on this.

PR #523, opened 2024-04-11
Fix for running automation test on the tester machine

byteocean:
It seems to be the most probable fix for #513. As it does not fail for continues re-running of the corresponding build&test step and after the forcing-updating. And it makes dpdk app to give back huge page resource to OS after each execution. As explained in the [documentation](https://doc.dpdk.org/guides/linux_gsg/linux_eal_parameters.html), using this flag makes it impossible to start a secondary process, which is not required in automation tests. It is worthy of merging it to verify in a long term period.

```
root@dpdk-github-runner:/home/tli# grep HugePages /proc/meminfo
AnonHugePages:    223232 kB
ShmemHugePages:        0 kB
FileHugePages:         0 kB
HugePages_Total:       4
HugePages_Free:        3
HugePages_Rsvd:        0
HugePages_Surp:        0
root@dpdk-github-runner:/home/tli# grep HugePages /proc/meminfo
AnonHugePages:    202752 kB
ShmemHugePages:        0 kB
FileHugePages:         0 kB
HugePages_Total:       4
HugePages_Free:        4
HugePages_Rsvd:        0
HugePages_Surp:        0
```

### DIFF
--- a/test/dp_service.py
+++ b/test/dp_service.py
@@ -36,7 +36,7 @@ class DpService:
 			script_path = os.path.dirname(os.path.abspath(__file__))
 			self.cmd = f"gdb -x {script_path}/gdbinit --args "
 
-		self.cmd += f'{self.build_path}/src/dpservice-bin -l 0,1 --log-level=user*:8'
+		self.cmd += f'{self.build_path}/src/dpservice-bin -l 0,1 --log-level=user*:8 --huge-unlink '
 		if not self.hardware:
 			self.cmd += (f' --no-pci'
 						 f' --vdev={PF0.pci},iface={PF0.tap},mac="{PF0.mac}"'


PlagueCZ:
Makes sense that the flag would somehow help here, though the DPDK docs are pretty sparse. What does actually "no multiprocess support" mean? That `dpservice-dump` would not work? I would assume that only the process using this argument would be unable to map used pages, i.e. only if the secondary process would use it it would not work.

But as we need a fix for the issue anyway, I guess there is no other way right now and `dpservice,py` is only used in tests. At leas I personally run dpservice from the command-line directly and not via `dpservice.py` when I'm running real traffic.

guvenc (approved):
We will try this in main branch.

PR #526, opened 2024-04-16
`dp_ip_address` structure safety

### LINKED ISSUE #525: Dual IP address structure safety
Since the introduction of IPv6 overlay, there is a structure to hold either IPv4 or IPv6. This currently is not always set properly to the right type and even when it is, it's really easy to forget to in the future.

It would be best to have some wrappers to enforce proper type to be set when setting the value.

There is also a problem with the fact that it is being stored as a `union` which can be initialized improperly.

And IPv6 data is stored as a plain `uint8_t` array, so proper overflow checks against the size of this array are not always easy to do.

PlagueCZ:
I have improved the `struct dp_ip_address` to be "read-only" via `.ipv6`, `.ipv4` and `.is_v6`. It is now only writable using macros (inlines are not currently possible due to the need to `static_assert` the size of IPv6 array, there is a possibility of a followup PR to fix this). For example `DP_SET_IPADDR6(the_struct, ul_ipv6);`

The only drawback of this "read-only" behavior is that the usual structure assignment will fail (`*struct_x = *struct_y) and `rte_memcpy()` must be used. This is hidden by an inline function for `dp_ip_address`, but if you need to include it in some hash key, the whole key needs to also use `rte_memcpy()`

The second improvement closely connected to this (thus a single PR) is the use of `dp_ip_address` in hash keys as it is also packed now. This and the provided wrappers now ensure that the hash key is properly initialized and no "random" behavior should be possible (this has been achieved previously by complex code).

Due to the amount of code connected to this structure, I created a new `dp_ipaddr.h` and moved a few lines closely connected to it from various places.

The changes in test suite are only needed due to the fact that hash-keys are now structure differently (e.g. to optimize alignment), thus WCMP has changed.

(checkpatch fails for CPP header)

Fixes #525

### DIFF
--- a/include/dp_firewall.h
+++ b/include/dp_firewall.h
@@ -7,10 +7,13 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
+
 #include <sys/queue.h>
 #include <rte_common.h>
+#include "dp_ipaddr.h"
 #include "dp_mbuf_dyn.h"
-#include "dp_util.h"
+
+#define DP_FIREWALL_ID_MAX_LEN	64
 
 #define DP_FWALL_MATCH_ANY_PORT			0xFFFFFFFF
 #define DP_FWALL_MATCH_ANY_ICMP_TYPE	0xFFFFFFFF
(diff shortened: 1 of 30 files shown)


guvenc (approved):
Thanks Jay ! Looks good to me.

PR #529, opened 2024-04-18
Properly handle info of icmp packets in NAT port allocation

### LINKED ISSUE #527: Resource leak in netnat_portoverload_tbl_key table
**Describe the bug**
A port allocated in NAT context for an ICMP packet in `dp_allocate_network_snat_port` does not seem to get 
deleted with the corresponding function `dp_remove_network_snat_port` when the flow ages out.
This is how the key looks like in the functions mentioned below.

{nat_ip = 0xac150101, nat_port = 0x65, dst_ip = 0x2d5606ce, dst_port = 0x800, l4_type = 0x1}
{nat_ip = 0xac150101, nat_port = 0x65, dst_ip = 0x2d5606ce, dst_port = 0x40, l4_type = 0x1}

**To Reproduce**
Run the test ` test_vf_to_pf_network_nat_icmp` in debugger and set breakpoint to the functions mentioned above and 
observe the key for the table `netnat_portoverload_tbl_key`

More discussion here also: https://github.com/ironcore-dev/dpservice/pull/521

byteocean:
Fixes #527. icmp packet's identifier shall be properly filled into relevant key structures for NAT tables.

### DIFF
--- a/src/dp_nat.c
+++ b/src/dp_nat.c
@@ -596,7 +596,7 @@ const uint8_t *dp_lookup_network_nat_underlay_ip(struct dp_flow *df)
 
 	dst_ip = ntohl(df->dst.dst_addr);
 	dst_vni = df->tun_info.dst_vni;
-	if (df->l4_type == IPPROTO_ICMP)
+	if (df->l4_type == IPPROTO_ICMP || df->l4_type == IPPROTO_ICMPV6)
 		dst_port = ntohs(df->l4_info.icmp_field.icmp_identifier);
 	else
 		dst_port = ntohs(df->l4_info.trans_port.dst_port);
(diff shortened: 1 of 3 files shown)


PlagueCZ (changes requested):
This now fully fixes `test_vf_to_pf_network_nat_icmp` but to support NAT64, you need to add  `|| df->l4_type == IPPROTO_ICMPV6` to all three conditions.

In `main` it does not work for `test_vf_to_pf_network_nat_icmpv6` due to another bug, that is fixed in my branch, but I think it should be added here.

PlagueCZ (approved):
Looks good, fixes all reported problems

guvenc (approved):
Thanks

PR #537, opened 2024-04-29
Remove flows for a loadbalancer target on deletion

### LINKED ISSUE #534: Deleting loadbalancer targets does not remove connection-tracked flows for them
If a loadbalancer selects a target IP address, that target is then removed from the loadbalancer, connection tracking will still route packets to the (now removed) target instead of selecting a new one.

There is a test case already prepared in the new tests provided #533, just commented out to not fail automated tests.

PlagueCZ:
Similar to other instances (NAT, VIP, neighNAT, VM), there is now a call to prune the flow table.

I chose to simply match the flows using the target's underlay address, which should be unique.

The original idea was to find the right VNI, IP, protocol and port, but that requires not only one hash-table lookup (not a problem in gRPC), but also to match all loadbalanced ports (there can be multiple), etc, only to arrive at the same result.

But I may have overlooked some important detail, as I do not have the connection tracking and flow tables code really thoroughly in my head.

Fixes #534


byteocean (approved):
I see no issue in this approach, especially using underlay_dst as the filter criterial.

guvenc (approved):
Looks good to me as well.

Issue #541, opened 2024-05-08
IPv6 address type safety


PlagueCZ:
Currently all IPv6 addresses are implemented as `uint8_t[16]` array. This is problematic when passing them as arguments, because `sizeof(arg)` where function argument is `uint8_t *` or even `uint8_t[16]` is still only the width of the pointer type, not actually 16 bytes.

Also, copying/assigning IPv6 addresses is doe via `rte_memcpy(dest, src, size)`. Not only the size should be properly checked to be the same between source and destination, but also the call is actually doing a function call, condition to determine the length (and whether an optimized copy can be used) and then another call to the copy function itself which is a loop over a counter.

Having a custom implementation where only two simple `uint64_t` assignments are done (because that is 8+8 bytes) is much much faster.

This requires a simple, but extensive change to the codebase.

PR #543, opened 2024-05-08
Implement a custom type for IPv6

PlagueCZ:
Based on #541, all uses of IPv6 are now represented by a custom `union` which provides type-safety. Operations are done via inline wrappers that use optimized memory access.

Since the change affects the whole codebase, I was able to fund some small connected issues that are also included here (as the same lines would have to be changed twice if the PR got separated), but separate issues have been created and marked fixed by the appropriate commit.

Codecheck fails because of C++ and it also is not ready for `TAILQ` in structures.

### DIFF
--- a/include/dp_flow.h
+++ b/include/dp_flow.h
@@ -87,11 +87,9 @@ struct flow_nf_info {
 	uint32_t vni;
 	uint16_t icmp_err_ip_cksum;
 	enum dp_flow_nat_type nat_type;
-	uint8_t underlay_dst[16];
+	union dp_ipv6 underlay_dst;
 	uint8_t l4_type;
-} __rte_packed;
-static_assert(sizeof(((struct flow_nf_info *)0)->nat_type) == 1,
-			  "enum dp_flow_nat_type is unnecessarily big");
+};
 
 struct flow_value {
 	struct flow_key	flow_key[DP_FLOW_DIR_CAPACITY];
@@ -151,7 +149,7 @@ void dp_free_network_nat_port(const struct flow_value *cntrack);
 void dp_remove_nat_flows(uint16_t port_id, enum dp_flow_nat_type nat_type);
 void dp_remove_neighnat_flows(uint32_t ipv4, uint32_t vni, uint16_t min_port, uint16_t max_port);
 void dp_remove_iface_flows(uint16_t port_id, uint32_t ipv4, uint32_t vni);
-void dp_remove_lbtarget_flows(const uint8_t *ul_addr);
+void dp_remove_lbtarget_flows(const union dp_ipv6 *ul_addr);
 
 hash_sig_t dp_get_conntrack_flow_hash_value(const struct flow_key *key);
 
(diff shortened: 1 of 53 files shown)


include/dp_flow.h
```diff
@@ -87,11 +87,9 @@ struct flow_nf_info {
 	uint32_t vni;
 	uint16_t icmp_err_ip_cksum;
 	enum dp_flow_nat_type nat_type;
-	uint8_t underlay_dst[16];
+	union dp_ipv6 underlay_dst;
 	uint8_t l4_type;
-} __rte_packed;
-static_assert(sizeof(((struct flow_nf_info *)0)->nat_type) == 1,
-			  "enum dp_flow_nat_type is unnecessarily big");
```
PlagueCZ, line 94: Packing removed because the only use of this structure is in `struct flow_value` which is not packed.

include/dp_log.h
```diff
@@ -33,7 +33,7 @@ extern "C" {
 #define _DP_LOG_INT(KEY, VALUE) KEY, _DP_LOG_FMT_INT, VALUE
 #define _DP_LOG_UINT(KEY, VALUE) KEY, _DP_LOG_FMT_UINT, VALUE
 #define _DP_LOG_IPV4(KEY, VALUE) KEY, _DP_LOG_FMT_IPV4, VALUE
-#define _DP_LOG_IPV6(KEY, VALUE) KEY, _DP_LOG_FMT_IPV6, VALUE
```
PlagueCZ, line 36: This is done to force the compiler to check the type. Previously it was possible to pass any pointer to `DP_LOG_IPV6()` and the logger would then read 16 bytes from the stack, without care.

Now the compiler will stop you from passing anything else than a union/structure with a member `bytes`, which is still not perfect, but the best I could do in this crazy macro setup for structured logging.

PlagueCZ:
Tested the new code on a small OSC setup with garden clusters and shoots running on this version (bar the needed OSC patches), seems to be working.

guvenc (commented):
Thanks Jay. I am sure, this was not an easy refactor touching a lot of places but it is very well done. 
I just have one single comment. Otherwise looks good to me.

src/nodes/ipv6_nd_node.c
```diff
@@ -55,37 +52,39 @@ static __rte_always_inline rte_edge_t get_next_index(__rte_unused struct rte_nod
...
-	rte_memcpy(req_ipv6_hdr->src_addr, rt_ip, sizeof(req_ipv6_hdr->src_addr));
+	dp_set_src_ipv6(req_ipv6_hdr, gw_ip);
 
 	if (icmp_type != NDISC_NEIGHBOUR_SOLICITATION && icmp_type != NDISC_ROUTER_SOLICITATION)
 		return IPV6_ND_NEXT_DROP;
 
 	if (icmp_type == NDISC_NEIGHBOUR_SOLICITATION) {
 		nd_msg = (struct nd_msg *)(req_ipv6_hdr + 1);
-		if (memcmp(&nd_msg->target, rt_ip, sizeof(nd_msg->target)))
+		static_assert(sizeof(nd_msg->target) == sizeof(gw_ip->bytes), "Incompatible IPv6 format in ND message structure");
```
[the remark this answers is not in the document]
PlagueCZ: Exactly, it's just a paranoia check (or maybe Rust-like check?) since the `memcmp` comparison is "unchecked" in nature, so at least something is there and should be a compile-time thing, so why not just add it.

guvenc (approved):

Issue #497, opened 2024-02-14
Add loadbalancer listing to gRPC


PlagueCZ:
Currently there is no way to know which load balancers (loadbalancer IDs) are registered in dpservice.

Creating a `dpservice-cli list loadbalancers` would complete the API and provide that feature.

Rationale: when there is a problem in deployment and k8s/metalbond/metalnet/dpservice are out of sync, there is no easy way to recover from that without restarting

PR #550, opened 2024-05-21
Add the ability for gRPC to list loadbalancers

PlagueCZ:
As discussed in #497 this PR adds a gRPC call to list registered loadbalancers.

While creating the test suite to implement this, I noticed a few shortcomings in the gRPC testing module, so I reworked the pytest for it.

This PR is trying to be backward-compatible, thus there is a new call and a new message, but the message is not used in all places it should be. A `LoadBalancer` messge should be used in GET and CREATE messages too, but that would mean breaking the protocol.

If improving the protocol is preferred though, I can add another commit to this PR to rework the two APIs to also use the new message type. This would of course need an update to `metalnet` which currently is unaffected by this PR.

---
Pytest for this new call is in place and is tested to be working on my machine, but is not enabled at this time, because to make pytest work, it requires an updated `dpservice-cli` client, which can only be made available once this PR is merged (dues to the protocl definition being part of it).
Therefore a simple uncommenting PR will be required after the client is merged.


byteocean (changes requested):
Hi, tests show that this enhancement works without problem to list created LBs. Just one small point: the section `Create/delete/list loadbalancers:` in documentation also needs to be updated to reflect this enhancement.

byteocean (approved):

guvenc (approved):
Looks good. Thanks also for refactoring the test scripts.

PR #552, opened 2024-05-22
Add dpservice-cli and dpservice-go as sub projects under go and cli directories

guvenc:
This is a proposal for further discussion:

I added the dpservice-cli and dpservice-go current main states to the tools sub-directory of dpservice.
dpservice-cli has received a meson wrapper to its Makefile so that dpservice-cli can be built in the top
level meson using the following. (Resulting binary will be copied to top level build directory of dpservice)

`meson build -Dbuild_dpservice_cli=true`

without the option, the build behaves as before. This part already works in this PR.

dpservice-go can be kept separate and its go binding generation can be called only if needed in its own 
directory. (No need for meson integration ?)

Missing parts:
- github automated tests of dpservice-cli needs to be adjusted. (Test needs to be run in github action of dpservice)
- includes referencing the old url of the dpservice-go and dpservice-cli need to be adjusted. (metalnet ?)
- Checkpatch should not check the go files.
- Dockerfile needs to be adjusted. (To directly take the dpservice-cli from dpservice build directory)
- Version used in dpservice-cli should now use the same version of dpservice.


PlagueCZ:
This now seems to finished.
 - automated tests for `dpservice-go` have been added to the `test` action (as the library test requires running dpservice)
 - `dpservice-cli` now directly uses local library as it never actually needs to be built as a separate tool
 - Checkpatch actually only complained about spaces at the end of MarkDown lines, there is no need for filtering. Also current use of the script is ran on a diff output and not on files, thus we have no way of filtering by extension
 - Dockerfile now directly uses `dpservice-cli` built as a part of the build
 - tester image checks is the `dpservice-cli` version matches `dpservice-bin` version
 - the build of `dpservice-cli` is optional, you need to provide `-Dbuild_dpservice_cli` or `-Denable_tests` for it to be built alongside of dpservice itself
 - `meson` checks for the presence of `make` and `go` commands, but does not check any versions, which can be a problem for Go as we require at least 1.20, but in that case golang itself prints an error, and there is no good support for golang in meson, so I left it at that.

Metalnet will require updating, but can be done separately as there is no protocol change

PlagueCZ (approved):
Please note that I did not really "check" all files as most of them are copies from other repos. I did check all the changed files in dpservice repo.

guvenc:
Thanks guys @PlagueCZ @vlorinc 

This is already in pretty good shape. Just one comment:
- Why do I get this deprecation warning, when I compile ? Can we avoid this ?

image

PlagueCZ:
@guvenc iteresting, I do not see that warning
```
The Meson build system
Version: 1.0.1
```
It is indeed in the documentation, so I changed it, build works as previously for me.

.github/workflows/publish-docker-tester.yml
```diff
@@ -97,7 +97,34 @@ jobs:
...
           docker run --rm --privileged --mount type=bind,source=/dev/hugepages,target=/dev/hugepages ghcr.io/ironcore-dev/dpservice-tester:${{ env.DOCKER_IMAGE_SHA }}
+      - name: Docker cleanup
+        if: always()
+        run: |
+          for i in $(docker ps -q); do docker stop $i; done
+      # dpservice-go tests
+      - name: Run checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
guvenc, line 113: We need to use `${{ env.DOCKER_IMAGE_SHA }}` as ref here. otherwise it will not check out ok when we build for the main branch. (when the trigger of the action is not "pull_request")
@vlorinc @PlagueCZ

guvenc:
@PlagueCZ 
Now looks good. I have an old meson on my system:
Looks like the old version is more talkative.

```
The Meson build system                                       
Version: 0.53.2  
```

.github/workflows/publish-docker-tester.yml
vlorinc: It's fixed now and tests were successful.

PlagueCZ:
Meson builds dpservice-cli always (because I simply have to run the `make` everytime) and that is always in the output:
```
$ ninja -C build/
ninja: Entering directory `build/'
[1/1] Generating cli/dpservice-cli/dpservice-cli with a custom command
```
The long output is present only when you somehow delete everything, so I did not suppress anything else.

And by definition all paths printed by meson are always based on the "build" directory specified by running `meson setup`, so if I do some changes to the service, even that path is emitted by meson:
```
$ ninja -C build/
ninja: Entering directory `build/'
[3/4] Generating src/dpservice-bin with a custom command
```

guvenc:
@byteocean  looks like you can also get some verbosity using `ninja -C builddir -d explain`
so this should be good enough for the beginning. (and in that case you can clearly see that dpservice-cli is being built)

guvenc:
The contents of this branch is fully merged to main branch using tools outside of github.

PR #555, opened 2024-05-27
Fix version template to re-generate during build instead of configure phase

PlagueCZ:
Only after merging dpservice-cli and creating a test for proper version matching with dpservice-bin, a flaw got noticed.

Currently dpservice uses Git commit ID as version, but does so only during configure phase. Which is fine for CICD, but when developing, this version almost never gets updated when switching branches.

The client correctly applies the version every build, so now the test sometimes fails.

I changed meson to update VCS tags during build instead.


guvenc (approved):
Looks good to me.

PR #556, opened 2024-05-27
Make ND protocol neighbour advertisement flags RFC conform.

guvenc:
Solicited neighbour advertisement messages need to set router flag

As dpservice acts as a router on its VF ports. Router flag needs to be set to the solicited neighbour discovery messages as well.
https://datatracker.ietf.org/doc/html/rfc4861#section-7.2.4

Unsolicited neighbour advertisement hop limit was accidently filled. Neighbour advertisements do not have this field and filling this field sets some flags accidently in the neighbour advertisement packet.


PlagueCZ (approved):
Looks good based on bug description. I have no IPv6 overlay in lab environment to check OSC atm.

PR #560, opened 2024-05-29
Automated hardware and benchmarking tests

byteocean:
Automated hardware and benchmarking tests and its detailed usage and configuration can be found in the committed documentation part. Also small fix to the offloading code for issues found during the tests.


src/nodes/lb_node.c
```diff
@@ -57,6 +57,9 @@ static __rte_always_inline rte_edge_t get_next_index(__rte_unused struct rte_nod
 			    || df->l4_info.icmp_field.icmp_type == DP_ICMPV6_ECHO_REQUEST
 			) {
 				df->nat_type = DP_CHG_UL_DST_IP;
+				cntrack->offload_state.orig = DP_FLOW_OFFLOADED;
+				cntrack->offload_state.reply = DP_FLOW_OFFLOADED;
+				df->offload_state = DP_FLOW_NON_OFFLOAD;
```
PlagueCZ, line 62: If this (and the one-liner in `dnat_node.c`) are a bugfix they should have an issue and the commit message should mark it. Or this should be a separate PR

If it's not a bugfix and something only needed for test (does not seem like), then it should be a conditional compilation.

docs/testing/automated_benchmark_test.md
```diff
@@ -0,0 +1,110 @@
+# When to perform automated benchmarking tests
+Automated benchmarking tests are additional functional and performance tests to the existing TAP device-based ones. This test suit relies on a configured environment including hypervisors and started VMs, as well as configured SSH authorized key of the execution machine starting the benchmarking tests.
+This benchmarking testing suite is built upon and utilises 1) dpservice-cli to configure running dpservice instances or containers on hypervisors; 2) connectivity and performance testing script under the repository `hack/connectiviy_test`. In the end, running these benchmarking tests is useful for verifying if dpservice works correctly together with actual running VMs for both offloading and non-offloading modes. It also verifies if networking performance meets specified values during dpservice development.
+
+# Examplary command to invoke tests
+```
+./runtest.py --mode offload --stage cicd --docker-image ghcr.io/ironcore-dev/dpservice:sha-e9b4272 -v
+
+./runtest.py --mode both --stage dev -v
```
PlagueCZ, line 9: Just a thought for maybe easier operations, that we could rename the TAP script to `taptest.py` and this one to `vmtest.py` or something else, just to prevent confusion in the future?

docs/testing/automated_benchmark_test.md
```diff
@@ -0,0 +1,110 @@
...
+Alternatively, if this option is set as 'cicd', the above described docker image generating process will not happen. Instead, a docker image specified by the option "--docker-image" will be used on hypervisors. This is a required option.
+
+3. "--docker-image". This option specifies the container image to be deployed to hypervisors. It is optional but required for the 'cicd' stage.
+
+4. "--reboot". This option specifies if a reboot process needs to be performed on VMs. It is needed if test configurations have changed, e.g., private IP addresses of VMs, and VMs need to obtain new configurations. If you want to ensure a fresh start of VMs, this option can also be enabled. It is optional.
+
+5. "--env-config-file" and "--env-config-name". They provide information of the above described `test_configurations.json`. It is possible this file is renamed or located somewhere else. And it is also possible to create several configurations within this file and specify one of them for the tests.
+
+6. "--verbose". Specify if pytest runs in the verbose mode to see all steps and results during test execution.
```
PlagueCZ, line 110: Maybe use backticks for the parameters (here and the above sections), because the reuslting markdown with quotes is less readable (can be just me).

PlagueCZ (commented):
Looks quite big and thorough. I do not have this setup currently and I will be creating one in the near future but need another computer for that.
But since it already has big use-case for testing, that should not stop anything of course.

The only real thing to address for me is the marking of C code change by a simple `fixes XX`.

PlagueCZ (approved):
Looks good, I cannot do proper testing now, but will be in the future so we can update this afterwards

guvenc (changes requested):
I have successfully tested this script in my environment. Thanks a lot.
I have the following feedback:

1. Please make sure that licence information is added where it is needed to the new files.
2. Ctrl + C during init phase of the script wouldnt stop the script.
3. We need VM provisioning to make the automation, a real automation.
   -  An ideal VM provisioning would then get a pointer to a rsa key pair and vm xml file for each VM and xml file for each default network (virtio) of the hypervisor and together with an ignition file, it can provision the VMs. (with predefined usernames and automatic copy of the ssh keys via ignition)
 
I think second and third point above can be done as a second step. For the details of the third step, we can have a separate meeting to discuss. If the licence part is fixed, this PR can be seen as ready  to be merged, from my point of view.

guvenc (approved):
Looks good to me. Thanks.

PR #561, opened 2024-05-29
Remove 'Build and Publish Docker Image' from PR actions

PlagueCZ:
Removes an unnecessary and long "docker image build" action from PRs. The tester image already builds the same binary (along with others) and runs all tests.

The action will still be run for any push to main branch and for manual actions.

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -1,7 +1,7 @@
 name: Build and Publish Docker Image
 
 env:
-  platforms: ${{ (github.event_name == 'pull_request' && 'linux/amd64') || 'linux/amd64,linux/arm64' }}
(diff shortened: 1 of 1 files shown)


PlagueCZ:
The first part of the intended change does work, only the tester image has been built (and successfully tested).

As for the second part (still working for main), that remains to be seen. 

I have read only the documentation on the `on` keyword (not the whole "actions" documentation), which led me to simply remove the `on: pull_request` section to prevent the build on PRs. Then I saw multiple places testing whether or not the action has been run via this `pull_request` event, so I removed those too.

Hopefully I have not missed anything.

guvenc (approved):
Looks good to me. Let's see how it does on the main branch.

Issue #538, opened 2024-05-08
Add dpdk prometheus exporter testing to pytest framework


guvenc:
# Summary

Add dpdk prometheus exporter testing to pytest framework. This will ensure to catch combability issues as soon as possible,

PR #562, opened 2024-05-30
Move prometheus-dpdk-exporter into dpservice repository

PlagueCZ:
Similar to dpservice-cli and dpservice-go, this PS moves prometheus-dpdk-exporter into this repository.

As it is a client to telemetry socket, I moved it into `cli/` subdirectory. I also re-used the meson flag `build_dpservice_cli` as this simply now "builds all clients", but this can be changed if needed.

The docker image now builds the exporter locally without downloading anything and thus the whole Dockerfile for the original exporter is also gone and no new github action has been created to build it separately.

I also changed the name to `dpservice-exporter` which seems more fitting given the purpose and deep integration (kustomization has also been updated).

I skipped some of the files that are repository-based, like CODE_OF_CONDUCT and CONTRIBUTING as they seem to belogn to the root of the whole repo instead. This actually also applies to all the licenses, but seeing that dpservice-cli and dpservice-go also kept them, I think this should be cleaned-up in a separate PR (also some README files contain invalid REUSE links due to this).

As I needed to test this move locally, I also integrated issue #538 here (with some needed refactoring of duplicate code this caused). And to do that I slightly changed the `main()` function to properly provide `dpservice-exporter -h` and `dpservice-exporter -v`.


guvenc (approved):
Thanks. Looks good.

guvenc:
@PlagueCZ 
could you please rebase to main ?

PR #567, opened 2024-06-03
Create a script to read current version from git, fixes #566

### LINKED ISSUE #566: Wrong version string for commits with golang tag
After merging `dpservice-go` here, there is a need for tagging some commits with a version tag for the go library, e.g. `go/dpservice-go/v0.3.8-15-gfeb118f`.

Thus version "getters" for other components need to take this into account and only take the suffix `v0.3.8-15-gfeb118f`, not the whole output of `git describe --tags`

PlagueCZ:
Instead of manually calling `git describe --tags` in multiple places, there is a script that does it, so any changes are now centralized.

The script only takes the version suffix from git, i.e. everything after the last `/`

### DIFF
--- a/cli/dpservice-cli/Makefile
+++ b/cli/dpservice-cli/Makefile
@@ -62,7 +62,7 @@ test: fmt vet ## Run tests.
 
 .PHONY: build
 build: fmt vet ## Build binary.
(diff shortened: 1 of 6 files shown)


guvenc (approved):
Thanks for the quick solution.

PR #568, opened 2024-06-03
Add hash table fullness telemetry

### LINKED ISSUE #438: dpservice telemetry metrics to expose the internal table fullness
The most obvious and the important one is the fullness of the conntrack table but other tables which are used in dpservice can also have such metrics. Some suggestions would be:

- LB/DNAT Table fullness. 
- How many (vf) interfaces are actively used on that dp-service. (Interface table fullness)

CC: @PlagueCZ @dnlvanco

PlagueCZ:
Adds a general telemetry for all hash tables to monitor the saturation/fullness.

Due to the names of tables being visible in Prometheus output, most have been renamed based on a suggestion from Guvenc. In a few case the variables themselves got renamed, mostly due to the fact that they are no longer `ipv4_` only variables, since there is now an IPv6 overlay.

This PR also includes changes to the exporter and tests.

### DIFF
--- a/cli/dpservice-exporter/main.go
+++ b/cli/dpservice-exporter/main.go
@@ -63,6 +63,7 @@ func main() {
 	r.MustRegister(metrics.InterfaceStat)
 	r.MustRegister(metrics.CallCount)
 	r.MustRegister(metrics.HeapInfo)
+	r.MustRegister(metrics.HashTableSaturation)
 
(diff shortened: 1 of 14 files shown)


guvenc (approved):
Looks good. So nice to have these statistics. They will be valuable to operations.

guvenc:
@PlagueCZ 
Could you please rebase ?

PR #569, opened 2024-06-04
Move dp_grpc_client into cli/ subdirectory

PlagueCZ:
To make the original C client in-line with the new repository structure, I moved it to `cli/`.

For backward compatibility I let it compile always, but it could be changed easily if we want to only do it by some meson option.


guvenc (approved):
Makes sense to move it under cli/

PR #571, opened 2024-06-05
Add firewall telemetry - rule count

### LINKED ISSUE #549: dpservice temetry metrics to expose firewall rule count
Separating this issue from #438

As suggested by @guvenc:
It would also make sense to expose number of firewall rules installed per Interface. (even better if we could show the count for egress and ingress rules per interface)
Imagine a scenario where the customer complains about not reaching the vm from outside and we can see that there is no ingress firewall rule installed on that interface. (Because we "deny" all ingress traffic per default and we only have "allow" firewall rules.)
This can already help operations to see quickly, it is the firewall rule which was not installed to allow the incoming traffic.
```
{
  "/dp_service/firewall/rule_count": {
    "vm1": 0,
    "vm2": 0,
    "vm3": 0
  }
}
```

PlagueCZ:
Fixes #549 

We added a new interface stat to the exporter (and thus telemetry) to show number of firewall rules per interface.

The only thing to discuss is commit 937f27edbdacdf4e64fed4cb08f4e1db7838f1cf where I added a new field to the internal inteface structure. Without it, getting the count of the rules means iterating the `TAILQ` list.

As telemetry can be called multiple times per minute in practice, I decided to optimize this by creating a dedicated variable. If that is not wanted, I'll simply remove the commit.

---
This PR can be squashed, as all the commits belong together, but were split due to the above discussion and the fact that two people worked on it.


guvenc (approved):
I find the idea of caching the counter in interface structure totally fine.
Looks good to me.

PR #574, opened 2024-06-06
Fix DNAT cleanup in VNI reset

### LINKED ISSUE #573: VIP creation blocked by previously removed NAT
**The problem**
Sometimes in production, when a new cluster is to be put up (via VIP-based bootstrap VM), we get a DNAT_EXISTS error that prevents a VIP from being set-up, thus making the VM inaccessible.

**The cause**
After log analysis it seems that not all neighboring NAT entries are removed via gRPC, while the whole VM (NAT and interface) got removed. This causes the VNI data to be reset without fully cleaning up.

**Further information**
Confirmed by @guvenc in metalnet log analysis, the VNI gets unsubscribed before all neighbor removal messages are received. This is not a problem by design, just that the VNI cleanup needs to be fixed.

This bug only ever happens when a VIP needs to be created on a VNI previously occupied by a NAT that got removed in the way described above, as only VIP can end in DNAT_EXISTS error. This is also why it is only observable by OSC in the case of cluster bootstrapping.

**Proposed solution**
Add DNAT entry cleanup to VNI reset.

PlagueCZ:
This should be enough to fix the problem for VNI unsubscription and neighboring NAT removal

Fixes #573

### DIFF
--- a/src/dp_nat.c
+++ b/src/dp_nat.c
@@ -823,6 +823,8 @@ void dp_del_all_neigh_nat_entries_in_vni(uint32_t vni)
 	for (item = TAILQ_FIRST(&nat_headp); item != NULL; item = tmp_item) {
 		tmp_item = TAILQ_NEXT(item, entries);
 		if ((item->vni == vni) || (vni == DP_NETWORK_NAT_ALL_VNI)) {
+			// can fail on non-presence (fine) or internal error (already logged)
+			dp_del_dnat_ip(item->nat_ip, item->vni);
 			TAILQ_REMOVE(&nat_headp, item, entries);
 			rte_free(item);
 		}
--- a/src/dp_nat.c
+++ b/src/dp_nat.c
@@ -817,6 +817,28 @@ int dp_list_nat_neigh_entries(uint32_t nat_ip, struct dp_grpc_responder *respond
 	return DP_GRPC_OK;
 }
 
+static int dp_del_dnat_by_vni(uint32_t vni)
+{
+	struct dnat_data *data;
+	const struct nat_key *key;
+	uint32_t iter = 0;
+	int	ret;
+
+	while ((ret = rte_hash_iterate(ipv4_dnat_tbl, (const void **)&key, (void **)&data, &iter)) != -ENOENT) {
+		if (DP_FAILED(ret)) {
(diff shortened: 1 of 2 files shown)


PlagueCZ:
I added one more commit to optimize the DNAT entry deletion function to reduce hashing, since it will be called multliple times in VNI reset. Even though this is in orchestration phase, so no hot-path.

I think these commits can be squashed before merging, but for current analysis I kept them separate.

src/dp_nat.c
```diff
@@ -823,6 +824,8 @@ void dp_del_all_neigh_nat_entries_in_vni(uint32_t vni)
 	for (item = TAILQ_FIRST(&nat_headp); item != NULL; item = tmp_item) {
 		tmp_item = TAILQ_NEXT(item, entries);
 		if ((item->vni == vni) || (vni == DP_NETWORK_NAT_ALL_VNI)) {
+			// can fail on non-presence (fine) or internal error (already logged)
```
guvenc, line 827: I was actually thinking what if the neigh nat linked list is empty and we still have an entry in dnat table ?
but this can not happen because the way `dp_del_vip_from_dnat` is implemented ? or would it be safer to call smt like this from `dp_free_vni_data` and do not touch `dp_dell_all_neigh_nat_entries_in_vni`

Disclaimer: Code below never run or tested.

```
int dp_del_dnat_by_vni(uint32_t vni)
{
    const void *key;
    void *data;
    uint32_t next = 0;
    int ret;
    struct nat_key *nkey;

    while ((ret = rte_hash_iterate(ipv4_dnat_tbl, &key, &data, &next)) >= 0) {
		if (DP_FAILED(ret))
			return DP_GRPC_ERR_ITERATOR;

        nkey = (struct nat_key *)key;
        if (nkey->vni == vni) {
            rte_free(data);

            if (DP_FAILED(rte_hash_del_key(ipv4_dnat_tbl, key))) {
                DPS_LOG_WARNING("Failed to delete DNAT key for IP: %u, VNI: %u\n", nkey->ip, nkey->vni);
            }
        }
    }

        return DP_GRPC_OK;
}

```

guvenc: The last time we discussed and looked at the details of the hashtable implementation. our conclusion was that it is safe to delete during iteration ? At least conntrack hashtable still does it that way. (and some other hashtables as well)

guvenc (approved):
Looks good now.

PR #575, opened 2024-06-10
Change gRPC request input error to a warning

PlagueCZ:
Just a tiny change to not print an error due to invalid gRPC input value. This is in line with all the other gRPC calls.


guvenc (approved):
LGTM

PR #582, opened 2024-06-20
Move local tests to a separate directory

PlagueCZ:
The addition of "benchmark_test" introduced a forced argument `--ignore` to the current testing suite, without it it does not work anymore. This is cumbersome for development and in the future this will potentially need more and more such switches when other tests are added.

Thus I moved the original suite into its own subdirectory.


PlagueCZ:
It just occurred to me that the tester docker image is in use somewhere. This change does change the default path to dpservice but the default path is not the same now for the tester image so unless `dp_service.py` is called with `--build path` in that case, this change can break the use. But I cannot remember where the use of it was exactly to test. Some github action I think

PlagueCZ:
Oh and now it occurred to me that the action for test is using `dp_service.py` and works (see above). The reason is that even though the path `/test/../../build/` does not really exist, it does resolve properly because the docker image filesystem root resolves `..` as itself (the root). So it actually works after this change.

guvenc (approved):
Thanks.

PR #584, opened 2024-06-24
Fix conntrack reference count causing double free

### LINKED ISSUE #583: dpservice local test fails on one of the dell machines
dpservice local test fails on one of the dell machines (dell-3). The GDB shows the following error stack trace:

```
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x000055555597edbc in dp_ref_dec (ref=0x11fffd52c8) at ../include/dp_refcount.h:35
#2  0x000055555597f37f in dp_age_out_flow (flow_val=0x11fffd5200) at ../src/dp_flow.c:411
#3  dp_process_aged_flows_non_offload () at ../src/dp_flow.c:437
#4  0x0000555555702023 in dp_process_event_flow_aging_msg (m=0x17936a9c0) at ../src/monitoring/dp_event.c:120
#5  0x0000555555715eac in dp_process_event_msg (m=0x17936a9c0) at ../src/monitoring/dp_monitoring.c:22
#6  0x00005555558ed005 in handle_nongraph_queues () at ../src/nodes/rx_periodic_node.c:48
#7  rx_periodic_node_process (graph=0x11ffffd9c0, node=0x11ffffe380, objs=0x11fffea140, nb_objs=0) at ../src/nodes/rx_periodic_node.c:74
#8  0x00005555559d3850 in __rte_node_process (graph=0x11ffffd9c0, node=0x11ffffe380) at /usr/local/include/rte_graph_worker_common.h:186
#9  rte_graph_walk_rtc (graph=0x11ffffd9c0) at /usr/local/include/rte_graph_model_rtc.h:42
#10 0x00005555559d351f in rte_graph_walk (graph=0x11ffffd9c0) at /usr/local/include/rte_graph_worker.h:38
#11 0x00005555559d30df in graph_main_loop (arg=0x0) at ../src/dpdk_layer.c:100
#12 0x00007ffff7ae542f in eal_thread_loop () from /usr/local/lib/x86_64-linux-gnu/librte_eal.so.24
#13 0x00007ffff7afbe26 in eal_worker_thread_loop () from /usr/local/lib/x86_64-linux-gnu/librte_eal.so.24
#14 0x00007ffff70a645c in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#15 0x00007ffff7126bbc in ?? () from /lib/x86_64-linux-gnu/libc.so.6
```

After a debugging session with @PlagueCZ, it is found out that swapping the order of `union dp_ipv6 underlay_dst` and `uint8_t l4_type` or make `struct flow_nf_info` packed can eliminate this error. The root cause still needs to be investigated.

guvenc:
In conntrack hashtable each conntrack entry has two keys pointing to it. (Flow 5-Tuple and Inverted Flow 5-Tuple)
Therefore a better approach to delete the conntrack entry is to reference count for each pointing hashtable key and 
delete the conntrack entry only when there arent any keys pointing to it.

Current approach tried to delete the both keys at once when the conntrack entry was supposed to be deleted. This caused double free under some circumstances.

Fixes #583 
Probably related to #513

### DIFF
--- a/src/dp_flow.c
+++ b/src/dp_flow.c
@@ -214,7 +214,7 @@ void dp_invert_flow_key(const struct flow_key *key /* in */, struct flow_key *in
 	}
 }
 
-static void dp_delete_flow_no_flush(const struct flow_key *key)
+void dp_delete_flow(const struct flow_key *key, struct flow_value *flow_val)
 {
 	int ret;
 
@@ -229,11 +229,7 @@ static void dp_delete_flow_no_flush(const struct flow_key *key)
 #ifdef ENABLE_PYTEST
 	dp_flow_log_key(key, "Successfully deleted an existing hash key");
 #endif
-}
-
-void dp_delete_flow(const struct flow_key *key)
-{
-	dp_delete_flow_no_flush(key);
+	dp_ref_dec(&flow_val->ref_count);
 	// removed a flow, purge the cache to be safe
 	// (could only remove this key from cache, but that would need matching, which takes time)
 	dp_cntrack_flush_cache();
@@ -268,8 +264,6 @@ void dp_free_flow(struct dp_ref *ref)
 	struct flow_value *cntrack = container_of(ref, struct flow_value, ref_count);
 
 	dp_free_network_nat_port(cntrack);
-	dp_delete_flow_no_flush(&cntrack->flow_key[DP_FLOW_DIR_ORG]);
-	dp_delete_flow_no_flush(&cntrack->flow_key[DP_FLOW_DIR_REPLY]);
 	dp_cntrack_flush_cache();
 
 	rte_free(cntrack);
@@ -402,12 +396,6 @@ static __rte_always_inline int dp_rte_flow_query_and_remove(struct flow_value *f
 
 }
 
-static __rte_always_inline void dp_age_out_flow(struct flow_value *flow_val)
-{
-	flow_val->aged = 1;
-	dp_ref_dec(&flow_val->ref_count);
-}
-
 void dp_process_aged_flows_non_offload(void)
 {
 	struct flow_value *flow_val = NULL;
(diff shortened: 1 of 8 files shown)


src/dp_flow.c
```diff
@@ -430,16 +423,16 @@ void dp_process_aged_flows_non_offload(void)
...
 	}
 }
 
-static __rte_always_inline void dp_remove_flow(struct flow_value *flow_val)
+static __rte_always_inline void dp_remove_flow(struct flow_value *flow_val, const void *key)
 {
 	if (offload_mode_enabled)
 		dp_rte_flow_remove(flow_val);
-	dp_age_out_flow(flow_val);
+	dp_delete_flow_no_flush(key, flow_val);
```
PlagueCZ, line 435: I think `no_flush` is only to be used if you do multiple deletions and then you flush manually (that's why `no_flush` is private/static here, flushing needs to happen after deletion otherwise there can be a problem with a cached pointer to the flow structure (that can be freed here).

This can be solved by calling `dp_cntrack_flush_cache()` after the hash-table iteration. But there is high probability that when we change something in one of these hashtable traversals, this will be overlooked and we can get an inconsistent cache state.

I would prefer to just drop the non-flushed version altogether as the only reason for it to exist previously was the code you just removed. I understand that this will be little bit less optimized in terms of performance, but it will be much safer. Especially compared to the previous state where the flushing happened also multiple times per iteration anyway.

PlagueCZ (changes requested):

src/dp_flow.c
guvenc: I am a bit confused what you mean here.
Do you suggest that we remove the "no_flush" from function naming of `dp_delete_flow_no_flush` ? or we delete the function  `dp_delete_flow_no_flush` completely ? in that case, there will be no function to delete the hashtable keys ?

PlagueCZ: I meant move the code from the no-flush version inside the flushing version and only use the flushing version

PlagueCZ (approved):
Looks like a great fix to the problem. Tests are also silent for me.

PR #585, opened 2024-06-27
Update documentation after repository merge

PlagueCZ:
As I was installing dpservice on a new computer I used the documentation as a guide and noticed small discrepancies (mostly about golang).

I also added a more verbose error concerning SR-IOV that I misconfigured, but dpservice was not helpful enough for me to get at the root cause.


guvenc (approved):
Thanks

PR #586, opened 2024-06-28
Do not run/fail gRPC GetInit test when attaching to a running service

PlagueCZ:
As noticed by those debugging the service over the last few weeks, pytest `test_grpc_getinit` fails when attaching to a running service. That's because the test actually checks the UUID of the service against the one obtained from running it. But when attaching to a running service, pytest was not the one to run it, thus has no UUID to compare to.

I made pytest simply skip this test when `--attach` is used.


guvenc (approved):
Thanks. Makes the life easier when testing locally.

PR #589, opened 2024-07-15
Introduce multiport-eswitch mode to support continuous communication

byteocean:
This PR introduced multiport-eswitch mode to support continues communication when one of the uplinks is down for reasons like maintenance. It includes:

1) a modified prepare.sh script to enable multiport-eswitch mode on host;
2) proper port initialisation using the async template approach that support multiport eswitch configuration;
3) a workaround solution that mitigate failed pf1 communication once this mode is enabled;
4) virtualservice implementation that uses the async template method and code enhancement for the above features (@PlagueCZ ).

To use prepare.sh script to configure the host and generate new dp_service.conf file, run `./prepare.sh  --multiport-eswitch --pf1-proxy`.

To run dpservice to coop with this multiport eswitch mode and use the workaround, run `sudo ./build/src/dpservice-bin -l 0,1 -n 2 -r 2 --   --no-stats --no-offload --multiport-eswitch`

A branch that also contains tap-based tests are also available, but it needs further discussion once tap-based workaround needs to stay longer.


hack/prepare.sh
```diff
@@ -171,33 +206,60 @@ function get_ipv6() {
...
-	echo "a-pf0 ${devs[0]},class=rxq_cqe_comp_en=0,rx_vec_en=1,representor=pf[0]vf[0-$[$actualvfs-1]]";
-	echo "a-pf1 ${devs[1]},class=rxq_cqe_comp_en=0,rx_vec_en=1"; } >> "$CONFIG"
+	echo "no-stats"
+	echo "pf0 $(get_ifname 0)"
+	echo "pf1 $(get_ifname 1)"
+	echo "vf-pattern $(get_pattern ${devs[0]})"
+	echo "ipv6 $(get_ipv6)"
+	if [[ "$OPT_MULTIPORT" == "true" ]]; then
+		echo "a-pf0 ${devs[0]},class=rxq_cqe_comp_en=0,rx_vec_en=1,dv_flow_en=2,dv_esw_en=1,fdb_def_rule_en=1,representor=pf[0-1]vf[0-$[$actualvfs-1]]"
+		if [[ "$OPT_PF1_PROXY" == "true" ]]; then
```
guvenc, line 230: There is no indication on the log output that PF1 proxy was enabled. Would be nice to have it in the logs.

hack/prepare.sh
```diff
@@ -9,6 +9,9 @@ set -Eeuo pipefail
 # Use as /usr/local/sbin/dp-prepare.sh
 #
 
+OPT_MULTIPORT=false
+OPT_PF1_PROXY=false
```
guvenc, line 13: Why dont we get these values as parameters to the script ?
Smt like this

```bash
OPT_MULTIPORT=false
OPT_PF1_PROXY=false

while [[ "$#" -gt 0 ]]; do
  case $1 in
    -multiport)
      OPT_MULTIPORT=true
      ;;
    -pf1proxy)
      OPT_PF1_PROXY=true
      ;;
    *)
      echo "Invalid option: $1" >&2
      exit 1
      ;;
  esac
  shift
done
```

This can be better consumed/set in the kubernetes yaml files configuring the dpservice init container using this script.

guvenc (changes requested):
Thanks Tao and Jay. I tested it functionally and looks good but 
I have some comments though and the following happened to me:

If I run the standard prepare.sh with enabled proxy_pf1 and multiport switch and run the dpservice-bin with generated config
I get this:
(My firmware setting has maximum possible VFs 126)

E SERVICE: Port id too high for Rx nodes, value: 128, max: 128 [../src/nodes/rx_node.c:37:rx_node_create()]

and dpservice exits.

src/nodes/cls_node.c
```diff
@@ -130,6 +167,11 @@ static __rte_always_inline rte_edge_t get_next_index(__rte_unused struct rte_nod
 	struct dp_virtsvc *virtsvc;
 #endif
 
+#ifdef ENABLE_PF1_PROXY
+	if (pf1_tap_proxy_forward(m))
```
guvenc, line 171: What about an unlikely here ?

PlagueCZ: Actually if anything shouldn't this be `likely`? If you are compiling with proxy, the only reason to do so is to run it (or test it but in that case optimization does not matter).

hack/prepare.sh
PlagueCZ: We do use these as a parameter of the prepare script exactly for the reason that I can then easily edit `kustomization.yaml` of the init container to switch to multiport. It also sets the `dpservice` command-line option `multiport-eswitch` inside `dp_service.conf` so there is no need to edit anything else.

Your code matches mine at the bottom of the file. This is only to set defaults, because we have the `set -u ` in place so bash complains about unset variables.

hack/prepare.sh
PlagueCZ: Good point. some equivalent to the already present "enabling multiport eswitch mode for $pf". The only "problem" is that the script actually doe not do anything with the tap device it only sets it to enabled and chooses the MAC address.
So I guess just add `enabling pf1 proxy using tap device $mac`?

PlagueCZ:
> Thanks Tao and Jay. I tested it functionally and looks good but I have some comments though and the following happened to me:
> 
> If I run the standard prepare.sh with enabled proxy_pf1 and multiport switch and run the dpservice-bin with generated config I get this: (My firmware setting has maximum possible VFs 126)
> 
> E SERVICE: Port id too high for Rx nodes, value: 128, max: 128 [../src/nodes/rx_node.c:37:rx_node_create()]
> 
> and dpservice exits.

Oh that would have bitten me in our lab testing. I forgot to test the high number.

This is because the TAP port is not counted the same way like the others, which is a deliberate choice by Tao, but made sense to me also as we handle it differently. As a quick fix for your test simply lowering the number of VFs (i think even by one) will make it work, but this needs further checking of course.

PlagueCZ:
When compiled with `enable_pf1_proxy`, the service will never run in the "normal mode" anymore, as the code does not actually check for the presence of the command-line switch. This is needed for our testing to be able to switch without actually changing the image.

hack/prepare.sh
guvenc: Sounds good.

hack/prepare.sh
guvenc: Yeah. Friday evening review. I didnt see the code below.

src/nodes/cls_node.c
guvenc: Well, this is classify node right ? When we enable proxy and we have 100 VMs running traffic on top of dpservice with their workloads and there is "one" PF1 interface where we check whether the traffic needs to be forwarded to TAP proxy port. At any given time in classify node, is it more likely to be a PF1 traffic that we classify or is it the traffic from one of the VF ports ?

PlagueCZ (approved):
This patch seems to solve the issue with switches going down causing loss of communication completely.
Tested on a local setup, pytest with mellanox patch (PR not ready) and in a lab cluster, but there only a partial test has been done.

In the future, if the PF1-proxy is to be kept (i.e. not conditional compilation anymore), some changes need to be done to make it better integrated.

guvenc (approved):

PR #600, opened 2024-08-26
Remove rollback from prepare.sh

### LINKED ISSUE #599: Prepare script gets stuck on rollback of numvfs
If setting eswitch mode fails, `prepare.sh` tries to set number of VFs to zero.

This is a problem if VMs are already up and running, because this action of setting to zero gets stuck (I think on waiting for all VMs to release the sriov VF device).

This happend on an OSC compute node when testing the multiport e-switch. Something was not working so the prepare script wanted to roll the number of VFs to zero, thus blocking the card till I manually evicted all VMs (and broke the stack in the process).

I think simply logging the error for the operator to fix it is better as then even on error, the rest of the machine stays configured the way it was before.

PlagueCZ:
I removed the setting of number of VFs to zero if prepare script encouters an error in setting the eswitch mode.

This way the error is logged (and visible in the appropriate container), but the machine itself is not changed if the number of VFs has already been set on startup.

Fixes #599

### DIFF
--- a/hack/prepare.sh
+++ b/hack/prepare.sh
@@ -121,8 +121,7 @@ process_switchdev_mode() {
 
 	log "enabling switchdev for $pf"
 	if ! devlink dev eswitch set pci/$pf mode switchdev; then
-		log "can't set eswitch mode, setting VFs to 0"
-		echo 0 > /sys/bus/pci/devices/$pf/sriov_numvfs
+		err "can't set eswitch mode"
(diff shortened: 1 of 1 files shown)


byteocean (approved):
looks reasonable to me, as if this step fails, dpservice should also not run with the eswitch mode.

guvenc (approved):
Looks good to me.

Issue #606, opened 2024-09-25
Prepare multiport-eswitch dpservice for OSC deployment


PlagueCZ:
This is an overarching issue to group PRs needed for proper dpservice deployment in multiport-eswitch mode in OSC environment.

PR #607, opened 2024-09-25
Add the ability to skip NIC setup in prepare script

PlagueCZ:
OSC has moved the actual HW setup part of the prepare script into startup scripts of k3s.

The init container however is still useful to properly configure dpservice based on the actual HW setup (like interface names etc.)

Thus there is the need for prepare script to only generate the config without changing HW setup.

Part of #606


guvenc (approved):
LGTM

PR #608, opened 2024-09-26
Remove Rx RSS

PlagueCZ:
When implementing pf1-proxy over MLX5-bound VF on PF1 (i.e. setting `sriov_numvfs` to `1` for PF1), maximum throughput on PF0 goes from 24.7Gbps to 12.5Gbps if there is also any VM-VM traffic. Such traffic is also 0.8Gbps.

By removing Rx RSS, PF0 throughput is back to maximum and VM-VM if done in parallel is up to 4Gbps.

Part of #606 

---
I have not made this part of the pf1-proxy PR since this change would be easy to miss and I think it should be tracked separately.


guvenc (approved):
LGTM

PR #609, opened 2024-09-26
Adjust packet mbuf size to reflect MTU

PlagueCZ:
Due to #606 we needed to increase the packet buffer size to adjust for Jumbo frames as that is what host-host communication is using. This in turn raised the required number of hugepages for dpservice pod from 4G to 16G on our compute nodes.

So instead I created two memory pools, one for 1500 MTU, the other for 9100 MTU.

In normal operation, dpservice only encounters 1500 MTU packets, yet the packet buffer size is set to `RTE_MBUF_DEFAULT_BUF_SIZE` (rte_mbuf.h):
```
#define RTE_MBUF_DEFAULT_DATAROOM       2048
#define RTE_MBUF_DEFAULT_BUF_SIZE       \
         (RTE_MBUF_DEFAULT_DATAROOM + RTE_PKTMBUF_HEADROOM)
```

So this PR lowers the buffer size to decrease memory requirements in normal mode and uses jumbo frames in pf1-proxy mode. Unfortunately `dpservice-dump` does not know which mode is active so it needs to use the bigger variant always, though it only uses a small ring buffer, so the allocation is not huge.

I have deployed the 1518 size on an OSC cluster and it has been running for two weeks now without visible problems.

The 9118 pool is tested in OSC, but not in the current state (i.e. using TAP device), I chose to do it this way to separate this PR from the big changes to the proxy.

### DIFF
--- a/include/dpdk_layer.h
+++ b/include/dpdk_layer.h
@@ -37,16 +37,25 @@ extern "C" {
 // there are three periodic messages (ARP, ND, ND-RA) that could be sent at once
 #define DP_PERIODIC_Q_SIZE	(DP_MAX_PORTS * 3)
 
-// 40Gb/s with 1500B packets means ~9M packets/s
-// assuming 0.1s delay in processing means ~900k mbufs needed
+// 40Gb/s with 1500B packets means ~3.3M packets/s
+// assuming 0.1s delay in processing means ~350k mbufs needed
 #ifdef ENABLE_PYTEST
 #define DP_MBUF_POOL_SIZE	(50*1024)
 #else
-#define DP_MBUF_POOL_SIZE	(900*1024)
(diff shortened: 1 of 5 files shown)


PlagueCZ:
Some hard data from prometheus exporter:
 - old dpservice (with 900k packet memory pool) had `HeapSize: 3G` and `AllocSize: 2.5G`
 - new dpservice (with 350k packet memory pool) has `HeapSize: 2G` and `AllocSize: 1.5G`
 - new dpservice with pf1-proxy (thus another jumbo pool): has `HeapSize: 3G` and `AllocSize: 2G`

guvenc (approved):
Nice that we could reduce the memory footprint of the dpservice.

PR #611, opened 2024-09-27
Fix exporter for different NUMA nodes

PlagueCZ:
Currently dpservice-exporter uses `/eal/heap_info,0` on all nodes. Some nodes in OSC require the use of `/eal/heap_info,1`, so this PR uses `/eal/heap_list` to ask for the right (NUMA node?) number.

I am not that versed in Go, so I simply "copied" the code for interfaces, therefore this code would not work when there are multiple `heap_info` entries. But for our deployment it should be fine.

Connected to #606


guvenc (approved):
LGTM

PR #612, opened 2024-10-01
Increase normal packet mempool member size

PlagueCZ:
This is a hotfix for the previous change to mempool sizes. I forgot about adding the IPv6 tunnel header because OSC uses MTU 1400 in its VMs.

I wanted to add a pytest for some maximum MTU, however (with tap devices at least) DPDK stops being able to push the packet to an interface sooner than the memory limit is reached.

This does however fix a problem using Mellanox with VMs having 1500 MTU, I have tested in my local setup.

### DIFF
--- a/include/dpdk_layer.h
+++ b/include/dpdk_layer.h
@@ -44,10 +44,12 @@ extern "C" {
 #else
 #define DP_MBUF_POOL_SIZE	(350*1024)
 #endif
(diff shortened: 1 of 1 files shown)


PlagueCZ:
Yes, this was a hotfix from yesterday that I already discussed with @guvenc as it breaks operation on SAP side. I would think that the bigger PR takes more time to approve, so it's also separate here.

guvenc (approved):
Thanks

PR #613, opened 2024-10-01
Fix/exporter port

### LINKED ISSUE #610: Dpservice crashes due to early telemetry in multiport-eswitch mode
This happened in OSC environment when using multiport-eswitch and `dpservice-exporter`. 

By asking for `/ethdev/xstats` during initialization, dpservice crashes.

Easily reproducible by running `while true; do echo "/ethdev/xstats,0" | dpdk-telemetry.py; sleep 1; done` and then starting dpservice.

Stack trace is showing corrupted stack trace on the telemetry thread, so I would think it's some sort of stack overflow. But there are no debug messages from telemetry to know more.

vlorinc:
# Proposed Changes

- add flag to change grpc port
- grpc tcp connection is now tested as first step
- removed retries of socket connection test, first connection is tried in loop in retryInterval until success
- retryInterval lowered to 5 seconds
- update shutdown process:
-- when connection to dpdk fails, process ends and doesn't retry to connect as before
-- this way it can be handled by kubernetes and shown in restart counter
-- added graceful shutdown of http server
- udpate error handling:
-- queryTelemetry function now returns error, this error indicates problems with dpdk telemetry socket and shuts down the program
-- connection testing functions merged together to clean up code
- check for uid to use correct dpdk telemetry socket

Fixes #610 Dpservice crashes due to early telemetry in multiport-eswitch mode:
Metrics are updated only after successful connection to grpc port.

### DIFF
--- a/cli/dpservice-exporter/main.go
+++ b/cli/dpservice-exporter/main.go
@@ -76,8 +76,10 @@ func main() {
 				log.Infof("Reconnecting to %s", metrics.SocketPath)
 				conn = connectToDpdkTelemetry(log)
 				log.Infof("Reconnected to %s", metrics.SocketPath)
+			} else {
+			    metrics.Update(conn, host, log)
 			}
-			metrics.Update(conn, host, log)
+
 			time.Sleep(time.Duration(pollIntervalFlag) * time.Second)
 		}
 	}()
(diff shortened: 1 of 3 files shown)


PlagueCZ (approved):
Covers everything needed and more.

Tested in OSC lab and based on Grafana output it is working right.

When dpservice pod gets restarted, exporter is now waiting for gRPC which is absolutely critical for multiport-eswitch due to a bug.

guvenc:
@vlorinc 
This one has a conflict with main branch now. Would you please rebase to main ?
I think the conflict comes from the commit "[Fix exporter for different NUMA nodes](https://github.com/ironcore-dev/dpservice/pull/613/commits/a7ce7908067d4d66cb465c73efd91c0f51d4208a)" 
which is also part of this PR and that commit is already merged to main.

guvenc (approved):
LGTM

PR #615, opened 2024-10-01
Acquire neighboring router MAC dynamically

PlagueCZ:
Currently dpservice acquires neighboring router MAC addresses during initialization.

Not only this is a problem if there is a port replaced on a switch (thus the MAC changes), but in multiport-eswitch with pf1-proxy, the MAC is taken from the proxy port. But without dpservice running, there are no neigbors registered (because the proxy interface is effectively dead at that time), so the MAC needs to be loaded dynamically after a while.

Thus I added a mechanism that in case of a missing neighboring router MAC, a timer is started and the retrieval is retried again later.

The timer has an exponential backoff, starting with 1s and doubled each time, until it reaches 60s.

This PR is best read commit-by commit, since that shows the evolution of all the changes.

Connected to #606 

---
There was also a need for the init container to fail when some of the values for `dp_service.conf` were not available. I lumped it together here as there were already changes to the prepare script, but this can be separated if needed.


PlagueCZ:
Closing due to pending commits from previous PR (something I do not remember working like this in the past)

PR #616, opened 2024-10-04
Export link status of PF interface

vlorinc:
# Proposed Changes

Adds link status to PF interface metrics.
Status is checked via `/ethdev/link_status` telemetry command and if ethdev is PF (not having string "representor" in name), link status is exported as `dpdk_ethdev_link_status` metric.

This PR is based on #613.


vlorinc:
New PR will be open that covers these changes and adds more.

PR #625, opened 2024-12-03
dpservice-cli: Fix JSON output

### LINKED ISSUE #624: dpservice-cli: add extended info to json output
When listing interfaces via dpservice-cli, extended info (using flag `--wide`) is shown only in table output (default) and not in json output (`-o json`).

Extended info (NAT and Virtual IP) related to these interfaces, should be shown also in json output.

vlorinc:
## Proposed Changes

Changed InterfaceSpec json tags to show nat and vip fields in json output if `--wide` flag is used.
If Interface has some NAT or Virtual IP associated with it, this will be shown in JSON/YAML output and not only in default table view.

Fixes #624


PlagueCZ (approved):

guvenc (approved):
Thanks

PR #629, opened 2024-12-10
Fix early TCP establishment

### LINKED ISSUE #628: TCP connection gets established too soon
In OSC production we get `conntrack_table` overflow after a few hours on some nodes. We were able to see that these nodes are being flooded by repeated SYN, SYNACK packets.

Looking at [`dp_cntrack.c:44`](https://github.com/ironcore-dev/dpservice/blob/6d9272ffeecca909df74d7eb4d066854738808c6/src/dp_cntrack.c#L44) in `dp_cntrack_tcp_state()` the TCP state machine is a simplification of the real thing. (This is also true for virtual services separately).

Looking at the code, when receiving SYN, SYNACK, SYN, SYNACK, the code reaches ESTABLISHED state due to missing failure edges from DP_FLOW_TCP_STATE_NEW_SYN and DP_FLOW_TCP_STATE_NEW_SYNACK.

I have created a pytest for this in `fix/early_tcp_establish`. Please run pytest thusly: `pytest -xv test_synattack.py -s |grep -i "flow timeout"`
```
2024-12-10 00:39:26.391 3(worker) W SERVICE:  flow timeout is now 30s [../src/dp_cntrack.c:134:dp_cntrack_set_timeout_tcp_flow()]
2024-12-10 00:39:26.531 3(worker) W SERVICE:  flow timeout is now 30s [../src/dp_cntrack.c:134:dp_cntrack_set_timeout_tcp_flow()]
2024-12-10 00:39:26.683 3(worker) W SERVICE:  flow timeout is now 30s [../src/dp_cntrack.c:134:dp_cntrack_set_timeout_tcp_flow()]
2024-12-10 00:39:26.815 3(worker) W SERVICE:  FLOW TIMEOUT IS NOW 24hrs [../src/dp_cntrack.c:130:dp_cntrack_set_timeout_tcp_flow()]
```
Above you can see the timeout changes after SYN, SYNACK, SYN, SYNACK respectively.

---
I am unsure of the solution. Of course the best one would be to implement the whole TCP state machine, but that requires looking at source/destination and watching the ACK and SEQ numbers.

My other solution would be to add transitions for the two states mentioned above and to go back to DP_FLOW_TCP_STATE_NONE where appropriate (or other needed states). But I would like to fist discuss that.

PlagueCZ:
Fixes #628


guvenc (approved):
LGTM

PR #630, opened 2024-12-17
Implement hash table inspector

### LINKED ISSUE #498: Create a tool to inspect internal dpservice tables
It would be helpful in some edge cases to inspect hash tables inside dpservice. 

Such a tool is easily implemented using DPDK's secondary process feature, it's just a question of providing the right resource name and knowing the data structure.

And example is already implemented in osc/lb_table_inspector branch

PlagueCZ:
I created another tool in `tools` called `dpservice-inspect`. It's another secondary process that reads shared memory and dumps the hash table entries.

Almost all tables can be inspected fully, with the exception of table of interfaces which uses static memory for the port structures instead of allocated hugepage memory. To inspect this table, dpservice would need to allocate the port table in advance.

I also did not implement virtual service table inspection as virtual services actually create one hash-table per virtual service with 64k entries (ports) and that is a completely different design than the current implementation. Moreover the data is also not shared in hugepage memory.

This tool has already been used to find bugs and is successfully running in OSC.

Fixes #498


byteocean (approved):
I played a bit this tool (of course not with millions of flows in the table ) and it works nicely. It looks good to me.

guvenc:
@PlagueCZ 
Would you please rebase to main ?

guvenc (approved):
Thanks. This is a powerful tool for deep debugging.

PR #631, opened 2024-12-18
Fix an inconsistent variable attribute.

byteocean:
This inconsistent variable attribute leads to compilation error (error: 'node' was marked unused but was used [-Werror,-Wused-but-marked-unused]). Just caught it accidentally.


PlagueCZ (approved):
Nice, not sure why nothing caught this before

guvenc (approved):

PR #633, opened 2025-01-08
added InterfaceID and VNI to prefix table view, …

### LINKED ISSUE #632: dpservice-cli: improve prefix overview
# Summary

Currently `prefixes` and `lbprefixes` can be listed without filter to show them all (`dpservice-cli list prefixes`).
In this case, it is not evident from table view, with which interface are these prefixes associated.

## Motivation

To have this info available, table view needs to be adjusted.

vlorinc:
# Proposed Changes

VNI field added to Prefix struct, it is populated based on associated Interface.
Added VNI and InterfaceID columns to prefixes and loadbalancer prefixes table view (they are also included in json output).
```
dpservice-cli list prefixes
 VNI  Prefix         UnderlayRoute    InterfaceID 
```
Commands are now case insensitive, e.g.: `dpservice-cli list LoadBalancers`

Fixes #632

### DIFF
--- a/cli/dpservice-cli/cmd/command.go
+++ b/cli/dpservice-cli/cmd/command.go
@@ -9,6 +9,8 @@ import (
 )
 
 func Command() *cobra.Command {
+	cobra.EnableCaseInsensitive = true
+
 	dpdkClientOptions := &DPDKClientOptions{}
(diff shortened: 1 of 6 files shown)


guvenc (approved):
Thanks. LGTM

PR #634, opened 2025-01-13
Remove pf1 proxy

### LINKED ISSUE #597: Rework pytest mellanox support
The current implementation of `pytest --hw` relies on a cross-wiring PF0 and PF1 together.

This not only removes the ability to test WCMP but also requires manual access to the NIC to switch cables.

It would be better to use a simple reflector on the "other side" to send packets from PFs back for `pytest` to be able to intercept them (as for Mellanox, packets emitted by PFs are not interceptable).

### LINKED ISSUE #635: Remove PF1-Proxy implementation
As the solution for non-working PF1 in multiport-eswitch has been found (missing kernel parameters CONFIG_MLX5_CLS_ACT and CONFIG_NET_TC_SKB_EXT on debian based distributions), the workaround is no longer needed.

PlagueCZ:
As the solution for non-working PF1 in multiport-eswitch has been found, the workaround is no longer needed.

All code is removed and documentation is updated with the mention of missing kernel option in Debian.

Pytest suite is working for on-hardware tests using two locally connected machines and when `--port-redundancy` option is used this automatically tests if PF1 is working properly otherwise scapy would not be able to see packets returned by the reflector on the other machine.

Fixes #635
Fixes #597


byteocean (approved):
compilation and tests worked well in my environment.

guvenc (approved):
Thanks. LGTM

PR #638, opened 2025-01-17
Add listing of all nat entries to gRPC protocol

### LINKED ISSUE #637: Add the ability to list all nat entries without the knowledge of NAT IP
Currently without knowing the NAT IP, there is no way to check for any remaining NAT entries when metalnet-dpservice get out of sync.

It would be nice to have `dpservice-cli list nats` which would simply not filter the entries by NAT IP

PlagueCZ:
Added the option to list all NAT entries by providing address 0.0.0.0 as the filter (the client does that automatically if no filter is given).

During implementation I also noticed a bug that is closely connected, so I included it here. Also there was an obsolete log line for listing the neighboring nats.

Fixes #637

The change required changing the protobuf specification, but I only added a field at the end, which should ensure backwards-compatibility and also enable old client code to use `dpservice-go` library without any issues.
That is why the names in the gRPC structure are looking strange.

This code is running on one node in OSC LAB, with an old metalnet, so backwards compatibility is tested.


guvenc (approved):
Thanks

PR #649, opened 2025-02-21
Make TAP devices usable in normal traffic scenarios

guvenc:
Make TAP devices usable in normal traffic scenarios and not only pytest based test scenarios where MAC addresses and checksums can be ignored if not directly effecting the testing scenario.

### DIFF
--- a/src/dp_netlink.c
+++ b/src/dp_netlink.c
@@ -13,6 +13,7 @@
 #include "dp_log.h"
 #include "dp_netlink.h"
 #include "dp_util.h"
+#include "dp_port.h"
 
 static int dp_read_neigh(struct nlmsghdr *nh, __u32 nll, struct rte_ether_addr *neigh,
 						 const struct rte_ether_addr *own_mac)
@@ -33,6 +34,9 @@ static int dp_read_neigh(struct nlmsghdr *nh, __u32 nll, struct rte_ether_addr *
 				if (rt_attr->rta_type == NDA_LLADDR)
 					memcpy(&neigh->addr_bytes, RTA_DATA(rt_attr), sizeof(neigh->addr_bytes));
 			}
+			// If it is a tap device, we make an exception with own MAC address check. FeBOX case
+			if (dp_conf_is_tap_mode())
+				return DP_OK;
 			if (!DP_MAC_EQUAL(own_mac, neigh))
 				return DP_OK;
 		}
--- a/include/dp_port.h
+++ b/include/dp_port.h
@@ -196,6 +196,12 @@ struct dp_port *dp_get_port_by_pf_index(uint16_t index)
 	return index < RTE_DIM(_dp_pf_ports) ? _dp_pf_ports[index] : NULL;
 }
 
+static __rte_always_inline
+bool dp_conf_is_tap(void)
+{
+	return dp_conf_get_nic_type() == DP_CONF_NIC_TYPE_TAP;
+}
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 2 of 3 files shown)


PlagueCZ:
Hmm lookoing at what this does, it seems to me it would be much better to simply ask for config and then either call `dp_get_pf_neigh_mac()` or use our own. 
Because this way we are calling "get neighbor" and then "oh what but in this special instance actually return something else and skip 99% of the code here).
There's even more space for commenting about it in dp_port.c where the call starts.

src/dp_port.c
```diff
@@ -226,6 +226,7 @@ static struct dp_port *dp_port_init_interface(uint16_t port_id, struct rte_eth_d
 	// oveflow check done by liming the number of calls to this function
 	port = _dp_ports.end++;
 	port->is_pf = is_pf;
+	port->is_tap = dp_conf_get_nic_type() == DP_CONF_NIC_TYPE_TAP;
```
PlagueCZ, line 229: This is basically a runtime-constant, so putting it inside the port structure seems strange to me. We could create `dp_conf_is_tap_mode()` that basically is an inline around the above comparison to make it easier to use.

src/dp_port.c
```diff
@@ -500,16 +501,20 @@ static void dp_acquire_neigh_mac(struct dp_port *port)
 	struct rte_ether_addr pf_neigh_mac = {0};
 	int ret;
 
-	if (DP_FAILED(dp_get_pf_neigh_mac(port->if_index, &pf_neigh_mac, &port->own_mac))) {
-		DPS_LOG_WARNING("No neighboring router, setting timer", DP_LOG_VALUE(port->neighmac_period), DP_LOG_PORT(port));
+	if (port->is_tap) {
+		rte_ether_addr_copy(&port->own_mac, &pf_neigh_mac);
```
PlagueCZ, line 505: Maybe a comment on why we are putting "false information" into neighboring mac?

Also is this FeBox specific? Because we are planning to use TAP mode but then connect the dpservice to a router, in which case this code would prevent proper working no?

Maybe simply create `DP_CONF_NIC_TYPE_TAP_FEBOX` or something similar? The name needs improving.

Or maybe leave this as-is and we can solve it on our side once we get to that problem.

PlagueCZ (commented):
If any of this PR was verified by pytest, please note that it is currently unable to detect some errors: #650 
Maybe rebase and test on some older commit first.

src/dp_port.c
guvenc: Also tested by rebasing on older commit without the pytest breaking change.

src/dp_netlink.c
```diff
@@ -33,6 +34,9 @@ static int dp_read_neigh(struct nlmsghdr *nh, __u32 nll, struct rte_ether_addr *
 				if (rt_attr->rta_type == NDA_LLADDR)
 					memcpy(&neigh->addr_bytes, RTA_DATA(rt_attr), sizeof(neigh->addr_bytes));
 			}
+			// If it is a tap device, we make an exception with own MAC address check. FeBOX case
+			if (dp_conf_is_tap_mode())
+				return DP_OK;
```
PlagueCZ, line 39: Ooh nice solution. So basically for TAP you accept the own mac here as you wanted and if we ever need a true neighbor mac it can be easily changed here, right?

PlagueCZ (approved):
This now looks really nice. The OSC virtual lab can be adapted once I have something to work with.

src/dp_netlink.c
guvenc: This would already accept true neighbor MAC. It just can not differentiate the "error case" anymore when it is in TAP mode. (Error case being "the router sending TAP device MAC as router MAC")

PR #652, opened 2025-03-14
Helper environment variables for tools in container

PlagueCZ:
Previous commit(s) introduced the ability to specify a different gRPC port and DPDK file-prefix for all tools.

But when dpservice is ran as a pod, these values are defined in the YAML. It is then a usability hurdle to specify these on the command line all the time while using dpservice tooling. 

This PR simply creates environment variables that can be set from the pod YAML.


guvenc:
@PlagueCZ  Please rebase. Thanks !

guvenc (approved):
LGTM

PR #654, opened 2025-03-17
Implement a gRPC health service

### LINKED ISSUE #653: Add a gRPC health service
It would be helpful for k8s to not periodically call `dpservice-cli get init` as a process and instead use gRPC health checking.

This would also enable `metalnet` to better react to dpservice going down, which currently is not ideal (there is some manual probing inside IIRC).

Looking at the offical docs: https://grpc.io/docs/guides/health-checking/
And also by writing a minimal client/server in golang to test it, it seems that it is capable of a permanent TCP connection with keepalives and there is no need for any manual probing/reporting.

In the PoC this is achived by built-in health service handling. However in library dpservice uses: https://grpc.github.io/grpc/cpp/md_doc_health-checking.html
This does not seem the case.

So the way forward would be to simply implement the linked protobuf specification and a needed C++ object.

PlagueCZ:
Fixes #653 

I implemented the service "from scratch" because I could not find a built-in service (unlike golang has). But it's not that big in the end.

One difference to golng implementation - golang uses TCP keepalive (packets of length 0), I was unable to find such option here (without overriding some methods to set it for all sockets by calling `setsockopt`), so I used `GRPC_ARG_KEEPALIVE_TIME_MS`. This is layer 7 though, so instead of two 0-length packets, this uses two packets of length 17 and an additional ACK.
I chose the same keepalive interval as observed by TCPdump for golang implementation, 15s.

This implementation uses `std::mutex` and `std::condition_variable` to prevent periodic sending of current status, simply only send a "change" in the value of status of the service (that's what the lock+cv is used for).

### DIFF
--- a/config/default/dp-service.yaml
+++ b/config/default/dp-service.yaml
@@ -57,11 +57,9 @@ spec:
         - mountPath: /var/run/dpdk/rte/
           name: socket-dir
         livenessProbe:
-          exec:
-            command:
-            - dpservice-cli
-            - get
-            - init
+          grpc:
+            port: 1337
+            service: ""
           initialDelaySeconds: 10
           periodSeconds: 10
       - name: prometheus-agent
--- a/include/grpc/dp_grpc_health.h
+++ b/include/grpc/dp_grpc_health.h
@@ -1,0 +1,25 @@
+// SPDX-FileCopyrightText: 2023 SAP SE or an SAP affiliate company and IronCore contributors
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef __INCLUDE_DP_GRPC_HEALTH_H__
+#define __INCLUDE_DP_GRPC_HEALTH_H__
+
+#include "../proto/health.grpc.pb.h"
+#include <mutex>
+#include <condition_variable>
+
+class HealthService final : public grpc::health::v1::Health::Service {
+private:
+	std::mutex status_mutex_;
+	std::condition_variable status_condition_;
+	grpc::health::v1::HealthCheckResponse::ServingStatus status_ = grpc::health::v1::HealthCheckResponse::NOT_SERVING;
+
+public:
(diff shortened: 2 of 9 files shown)


PlagueCZ:
This is now running in OSC LAB.
We have replaced the old liveness and startup probes:
```
livenessProbe:
        exec:
          command:
            - dpservice-cli
            - '--address=127.0.0.1:1338'
            - get
            - init
        initialDelaySeconds: 90
        timeoutSeconds: 3
        periodSeconds: 10
        successThreshold: 1
        failureThreshold: 10
      startupProbe:
        exec:
          command:
            - dpservice-cli
            - '--address=127.0.0.1:1338'
            - get
            - init
        initialDelaySeconds: 1
        timeoutSeconds: 3
        periodSeconds: 5
        successThreshold: 1
        failureThreshold: 60
```
with
```
livenessProbe:
        grpc:
          port: 1338
          service: ''
        initialDelaySeconds: 90
        timeoutSeconds: 3
        periodSeconds: 10
        successThreshold: 1
        failureThreshold: 10
```
Florin also changed the health check inside metalnet to use the gprc health service and it seems to be working.

Thus I think this feature is implemented correctly.

---
Of course checkpatch is complaining about C++ code, please ignore.

guvenc (approved):
I think this is quite good and can be merged and offers an equivalent functionality as before. (with a best practice approach)
One thing which was missing before and also with this implementation is that we can not detect the hang of the worker thread in dpservice. 
Some sort of timeout mechanism is needed in this busy loop to be able to detect this

> 	while (cq_->Next(&tag, &ok) && ok) {
		call = static_cast<BaseCall*>(tag);
		while (call->HandleRpc() == CallState::AWAIT_MSG) {
			// wait for response from worker
		};
	}

what do you think ? @PlagueCZ 
This is probably a separate thing.

PR #655, opened 2025-03-17
Disable total flow meter configuration

PlagueCZ:
As this feature is no longer working on current cards, it is polluting the logs with warnings and doubling the number of logs per interface creation.

Until solved by a different implementation, I would disable the total flow rate setting altogether.

---
I had to touch the `int err` because it would then go unused, so I used it by the public flow rate code block, that's why I also changed that one.


byteocean (approved):
LGTM. Thanks.

guvenc (approved):
LGTM

PR #657, opened 2025-03-19
dpservice-cli: fix for pytest unable to detect errors

### LINKED ISSUE #650: Pytest unable to detect errors
After the latest change to `dpservice-cli` pytest is unable to detect errors anymore.

If you use this test file (as a `test_broken.py` for example) and run it via `pytest test_broken.py`:
```
from helpers import *

def test_grpc_errors(prepare_ifaces, grpc_client):
        grpc_client.addvip(VM4.name, vip_vip)
```
You will see no errors and the test passes. But adding a VIP to a non-existent interface should fail.

Indeed `dpservice` fails:
```
2025-03-10 00:00:17.429 3(worker) W GRPC: Failed request, grpc_request: 14, grpc_error: 205, grpc_message: NO_VM [../src/grpc/dp_grpc_impl.c:1111:dp_process_request()]
```
(you can use `pytest -s test_broken.py` to see this output)

Reverting to `d80bc5a4fc7ff22b7b93720f1effd626e25b0e9b` makes pytest fail (as it should).

I think the problem is tied to the fact that pytest dpservice-cli handler is reacting to return code `2` in a way that is expecting a JSON error response, which is not present.

vlorinc:
# Proposed Changes

- Changes in error rendering: errors that are returned to main are parsed and rendered based on output flag, if json output was requested, error will now be rendered as json (using JsonError struct similar to other objects)
- Pytest was updated to reflect new output formats
- Pytest now also always checks for response from grpc client  
- Error object also includes "source" field that indicates where the error originated (client, grpc, server)

Fixes #650

### DIFF
--- a/cli/dpservice-cli/cmd/common.go
+++ b/cli/dpservice-cli/cmd/common.go
@@ -5,14 +5,18 @@ package cmd
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"net/netip"
 	"os"
+	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
+	"github.com/ghodss/yaml"
 	"github.com/ironcore-dev/dpservice/cli/dpservice-cli/renderer"
 	"github.com/ironcore-dev/dpservice/cli/dpservice-cli/sources"
 	"github.com/ironcore-dev/dpservice/go/dpservice-go/api"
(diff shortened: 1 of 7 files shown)


PlagueCZ (commented):
This seems to work really well, the new approach is better. But I was part of the change, so we need a fresh a pair of eyes now.

guvenc (approved):
LGTM. I also liked the fact that we added some tests which would catch if the errors get ignored in a similar way.

PR #664, opened 2025-04-08
Flush old packets when switching to active mode

### LINKED ISSUE #663: Duplicated old packets when switching HA dpservice
Currently when switching to a backup dpservice, it already has some packets in its Rx queue(s).
These packets however can be even seconds old, for example in a VM-VM communication:
```
# ping 192.168.123.11
PING 192.168.123.11 (192.168.123.11) 56(84) bytes of data.
64 bytes from 192.168.123.11: icmp_seq=1 ttl=64 time=0.332 ms
64 bytes from 192.168.123.11: icmp_seq=2 ttl=64 time=0.356 ms
64 bytes from 192.168.123.11: icmp_seq=3 ttl=64 time=0.334 ms
64 bytes from 192.168.123.11: icmp_seq=2 ttl=64 time=1502 ms (DUP!)
64 bytes from 192.168.123.11: icmp_seq=1 ttl=64 time=2529 ms (DUP!)
64 bytes from 192.168.123.11: icmp_seq=3 ttl=64 time=478 ms (DUP!)
64 bytes from 192.168.123.11: icmp_seq=2 ttl=64 time=1502 ms (DUP!)
64 bytes from 192.168.123.11: icmp_seq=1 ttl=64 time=2530 ms (DUP!)
64 bytes from 192.168.123.11: icmp_seq=3 ttl=64 time=478 ms (DUP!)
64 bytes from 192.168.123.11: icmp_seq=4 ttl=64 time=0.696 ms
64 bytes from 192.168.123.11: icmp_seq=5 ttl=64 time=0.371 ms
```
As you can see PING packets 1,2,3 were re-transmitted by both port 3 and port 2 (this communication is a direct VM-VM on these ports on one host).

Given the fact this is 1 second per packet, this is not an issue of a slow switchover which takes milliseconds to happen.

PlagueCZ:
When switching from backup mode to active mode, dpservice can note down the current timestamp on a NIC.

Then all packets older than this timestamp can be thrown away (per-port basis) and when the first packet that is not old arrives, processing can resume normally (packets that are present are in the right order, just old).

This requires the NIC to have timestamping offload capability (which Mellanox does), in cases there is none, timestamp will be set to zero and no flushing will take place.

Fixes #663

### DIFF
--- a/src/dp_port.c
+++ b/src/dp_port.c
@@ -99,6 +99,8 @@ static int dp_port_init_ethdev(struct dp_port *port, struct rte_eth_dev_info *de
 
 	/* Default config */
 	port_conf.txmode.offloads &= dev_info->tx_offload_capa;
+	if (dev_info->rx_offload_capa & RTE_ETH_RX_OFFLOAD_TIMESTAMP)
+		port_conf.rxmode.offloads |= RTE_ETH_RX_OFFLOAD_TIMESTAMP;
 
 	if (dp_conf_get_nic_type() == DP_CONF_NIC_TYPE_TAP || dp_conf_is_multiport_eswitch())
 		nr_hairpin_queues = 0;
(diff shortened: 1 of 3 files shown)


guvenc (approved):
LGTM

PR #666, opened 2025-04-08
Lower memory footprint and packet pool

### LINKED ISSUE #665: Better packet pool size computation
Currently the packet pool `mbuf_pool` size is computed based on [line throughput](https://github.com/ironcore-dev/dpservice/blob/dee896987ecec07d76ee981b444c743fc0403341/include/dpdk_layer.h#L36).

After analysing HA switchover, this does not seem to reflect how DPDK works.
If true, we should see the whole pool in need of flushing after switching as the backup dpservice should simply fill the pool after a while. Also then gRPC should stop working as the pool is exhausted.

Instead what seems to happen is:
 - DPDK uses a fixed amount of packets per active port per Rx queue as per `rte_eth_tx_queue_setup()` 
 - currently 1024 per port per queue
 - When Rx burst happens and packets are not trivially small, each packet allocates another chunk (in line with DPDK mbuf documentation) - currently 256 packet burst 
 - there are a few internal packet allocations in dpservice, like gRPC/service messages/etc.
 - as only one thread is ever active in a worker, only one burst of packets seems to travel through the whole graph and then is being sent out (and released), only then another burst can be worked with, thus only *one* burst can be active at a time

Based on the above information the pool size should actually be: `<number of ports> * <rx queue size> + <burst size>`. Of course some headroom is recommended to include some of the internal queues and to have  a little bit "just in case".

PlagueCZ:
Implemented the mentioned equation to compute packet pool size.

Tested in OSC with telementry monitoring available packets via `echo "/mempool/info,mbuf_pool" | dpdk-telemetry.py`.

Changes are in line with the issue description:
 - 1024 packets per active port are removed from the pool and only released after the port is closed by removing a VM
 - available packets oscillate by a few hundred there and back, consistent with bursts and caching and internal queues

Fixes #665

### DIFF
--- a/include/dpdk_layer.h
+++ b/include/dpdk_layer.h
@@ -33,13 +33,25 @@ extern "C" {
 // there are three periodic messages (ARP, ND, ND-RA) that could be sent at once
 #define DP_PERIODIC_Q_SIZE	(DP_MAX_PORTS * 3)
 
-// 40Gb/s with 1500B packets means ~3.3M packets/s
-// assuming 0.1s delay in processing means ~350k mbufs needed
+// While dpservice is processing RTE_GRAPH_BURST_SIZE, the NIC will store packets here
+// Seen recommendations to keep this at 2*RTE_GRAPH_BURST_SIZE or 4*RTE_GRAPH_BURST_SIZE
+#define DP_RX_QUEUE_SIZE (4 * RTE_GRAPH_BURST_SIZE)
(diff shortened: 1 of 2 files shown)


guvenc (approved):
I conquer to decrease the memory footprint without losing functionality based on test results and the calculation you presented.

Could you please also rebase to main ?

Issue #671, opened 2025-04-18
Simplify virtual service isolation


PlagueCZ:
In relation to #668 (migration to DPDK24), it has been discovered that virtual service isloation rules are unnecessarily complex, as they are comparing not only destination underlay address but also service address and port (TCP/UDP).

This was previously needed because the underlay address used was the node address, i.e. the /64 prefix.

Then it has been changed so that each virtual service generates it's own underlay address.

Thus it is only needed to create an isolation rule for such underlay address, everything else is already exclusively associated to it.

PR #672, opened 2025-04-18
Update DPDK to 24.11.1

### LINKED ISSUE #668: Update to DPDK 24.11
As the latest LTE version is 24.11, dpservice should use that. 

A preliminary branch is already prepared as `feature/dpdk24`. It's not final as the change to DPDK's IPv6 structures warrants a change to dpservice structures, but it's just a cosmetic change that remains to be finished.

## Problems
However there are currently two blockers for using this version in production

> I also created a `feature/dpdk25` (only a small change needed) to make sure this is not already fixed, but neither problem went away.

### Virtual service isolation rule async templates fail
For some reason isolation rule templates (not rules themselves) are failing. If I understand it correctly, DPDK is trying to find a "matcher" for our template, but cannot find a suitable one. Again, if I understand the code I traced correctly, we are trying to match both IPv6 source and TCP port, which means having a relatively large mask, which for some reason in new implementation DPDK does not support.

You do not need to use any virtual services, just compiling them in and starting dpservice will show the error.

### Slow port start
After switching from 23.11.3 to 24.11, each call to start a port takes around a second to succeed! This means even in my local testing with 2PFs and 4VFs, startup takes around 7 seconds to complete. This is unusable in production where this would mean over 120s of startup time.

This only happens on Mellanox, not TAP devices in testing.

## Next steps
The slowdown needs to be traced and solved, though I have only focused on the virtual services at first, where I already am at the end of my expertise currently with rte_flow templates. More research is needed.

PlagueCZ:
Providing necessary changes to build against DPDK 24.

This update introduced two problems at first:
 - virtual services unable to create isolation rules
 - slow port startup (up to 1s per port)

Both of them can be fixed by removing newly introduced validation code. This is part of this PR.

While no changes to virtual service isolation is then needed, I still created PR #671 based on this one to simplify them and remove the problem altogether regardless of the DPDK patch.

Aside from this, only small changes to the codebase were required.

Fixes #668


Dockerfile
```diff
@@ -57,11 +57,12 @@ ENV DPDK_DIR=/workspace/dpdk-stable-${DPDK_VER}
...
-&& patch -p1 -R < ../hack/dpdk_23_11_3_mtu.patch \
-&& patch -p1 < ../hack/dpdk_23_11_3_fdb_def_rule.patch \
-&& patch -p1 < ../hack/dpdk_23_11_log.patch \
-&& patch -p1 < ../hack/dpdk_23_11_telemetry_key.patch \
-&& patch -p1 < ../hack/dpdk_23_11_ethdev_conversion.patch
+&& patch -p1 -R < ../hack/dpdk_24_11_mtu.patch \
+&& patch -p1 < ../hack/dpdk_24_11_fdb_def_rule.patch \
+&& patch -p1 < ../hack/dpdk_24_11_log.patch \
+&& patch -p1 < ../hack/dpdk_24_11_telemetry_key.patch \
+&& patch -p1 < ../hack/dpdk_24_11_no_pattern_template_validation.patch \
```
guvenc, line 64: I thought that we agreed to make this optional with a docker parameter ? or do I miss smt here ?

PlagueCZ: That was when the change would mean changing functionality. I alter discovered that we can simply not validate without changing any functionality. Validation was only a feature introduced in 24, so this was never used before.

The build was meant to be there because the previous solution of mine would prevent anyone from using sync-flow-api, but the current solution does not prevent anything.

We can still make it a parameter of course, it's just that this is basically similar to the MTU patch now, removing some additional behavior that has been added in 24 but we do not actually need.

guvenc: Thanks for clarifying. If the sync-flow-api is not effected then this can stay without Docker parameter.

guvenc (approved):

PR #673, opened 2025-04-18
Simplify virtsvc isolation

### LINKED ISSUE #671: Simplify virtual service isolation

PlagueCZ:
Based on PR #672

This change is not necessary due to the removal of validation in PR #672, but makes the codebase read for a situation when the validation is re-introduced.

Also this simpler implementation is actually functionally equivalent to the previous one.

Fixes #671


guvenc:
PLease rebase to main. Thanks !

byteocean (approved):
LGTM. But I do not get the statement "...makes the codebase read for a situation when the validation is re-introduced".  Does it mean  'the codebase ready and working when the validation is enabled again'?

guvenc (approved):
LGTM too.

PlagueCZ:
`s/read/ready/` :)

PR #674, opened 2025-04-18
Remove soon to be obsolete __rte_packed

PlagueCZ:
DPDK 25 obsoleted `__rte_packed` in favor of a (more compatible with MSVC) pair of `__rte_packed_begin` and `__rte_packed_end` as per [this commit](https://github.com/DPDK/dpdk/commit/154303b06ab29e040000f0536961601d952e2974).

As a proposal of a solution I created this PR. The other solution would be to use both new macros, but they look worse than this solution, so basically this draft is a discussion PR.

### DIFF
--- a/include/dp_flow.h
+++ b/include/dp_flow.h
@@ -62,7 +62,7 @@ enum dp_flow_nat_type {
 	DP_FLOW_LB_TYPE_LOCAL_NEIGH_TRAFFIC,
 	DP_FLOW_LB_TYPE_RECIRC,
 	DP_FLOW_LB_TYPE_FORWARD,
-} __rte_packed;
+} __attribute__((__packed__));
 
 enum dp_flow_tcp_state {
 	DP_FLOW_TCP_STATE_NONE,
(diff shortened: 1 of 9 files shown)


guvenc (commented):
So the only drawback is that we would not be compilable with MSVC ?
Being compatible to MSVC would only cost us the readability ? I mean it is also not that bad we can see where the pack starts and ends ? or is there any other disadvantage then readability ?

PlagueCZ:
The disadvantage would be, that by using the new macros, we cannot easily switch between DPDK24 and DPDK 25, we would need to always have the transitional commit for DPDK 25.

Other than that the new macro is not bad or anything.
We could then maybe create a commit on DPDK 24 that defines the start and end in an ifdef:
```
#ifndef __rte_packed_begin
#define __rte_packed_begin
```
and so on
than way we can switch DPDK whenever too

guvenc:
Hmm. Can't we cherry pick this [commit](https://github.com/DPDK/dpdk/commit/154303b06ab29e040000f0536961601d952e2974) to DPDK 24.11 as part of the docker build process ? For DPDK 25, we can drop that cherry-pick ?

PlagueCZ:
Not exactly as we still need to actually define __rte_packed there, but I get the idea, just another patch to DPDK as usual, that is also a solution

guvenc:
Ok. I would suggest then that we start to use the new macro with start and end in dpservice as it is in DPDK 2025 and do a patch for 2024 DPDK with the ifndef as you suggested and that patch will be dropped when we switch to DPDK 2025.

PR #676, opened 2025-05-07
Fix and enhance benchmark test scripts.

### LINKED ISSUE #677: Automated benchmark tests are not compatible with enhanced dpservice
automated benchmark tests using real hardware and VMs fail during provision and testing, due to the enhancement of dpservice. To ensure this convenient benchmarking test working, they need to be patched.

byteocean:
Automated benchmark testing scripts need to be adjusted to several recent enhancement of dpservice. Thus, following changes have been made to make it working:

1) only docker-based tests for both dev and cicd (no need to run dpservice natively on hosts);
2) use dpservice-cli in container to ensure maximum compatibility (before dpservice-cli needs to be copied into local /tmp ); 
3) use google dns for VMs in any case to ensure a working apt installation in more cases where host configuration has flaws (for some reason, default dns didn't work on one of the hosts); 
4) adjust docs.

It should fix #677


guvenc (approved):
LGTM

PR #684, opened 2025-06-03
Add missing zmalloc check

### LINKED ISSUE #686: Fix several missing memory frees
Fix several missing memory frees

PlagueCZ:
There was a check missing after allocation. 

I also added a "rollback", though in that case I don't think there is any hope for recovery.

The rollback is now also done if cannot be added to a hash table, which actually can easily happen if it gets full, so please check that such rollback actually helps and not makes things worse.

While reading the code I did some simplifications too.

### DIFF
--- a/src/dp_nat.c
+++ b/src/dp_nat.c
@@ -690,7 +690,7 @@ int dp_allocate_network_snat_port(struct snat_data *snat_data, struct dp_flow *d
 
 	}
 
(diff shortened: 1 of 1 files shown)


byteocean (approved):
thanks for taking care of this rollback. LGTM

guvenc (approved):
LGTM as well.

PR #685, opened 2025-06-03
Add missing frees in argument parsing

### LINKED ISSUE #686: Fix several missing memory frees

PlagueCZ:
There were some edge cases where freeing of command-line argument memory was not properly done.

No real-world implication as this is just before the process exits anyway.

 - DHCPv6 DNS options were never freed
 - if `dp_service.conf` contains some options that allocate memory, then before actual command-line parsing the process fails, that memory would never be freed
 - if one mellanox option (in the config fileú does not properly allocate, the other mellanox options would not get freed

### DIFF
--- a/src/dp_conf.c
+++ b/src/dp_conf.c
@@ -327,6 +327,7 @@ int dp_conf_parse_file(const char *env_filename)
 void dp_conf_free(void)
 {
 	free(dhcp_dns.array);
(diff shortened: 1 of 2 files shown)


byteocean (approved):
thanks for fixing.

guvenc (approved):

PR #689, opened 2025-06-05
Add one tool for VM used in benchmarktest

byteocean:
Added nftable tool installation into the VM ignition file used in benchmark tests. It was missing probably due to rebasing error last time.


guvenc (approved):
LGTM

PR #690, opened 2025-06-16
Provide option to bind VFs to vfio driver

### LINKED ISSUE #680: Support `cloud-hypervisor` by extending init script
# Summary

In order to be able to consume the `VF`s provided by the `dpservice`, they must be bound to the `vfio-pci` driver. Since `dpservice` is owning the `VF`s it would make sense to prepare them there. 

## Basic example

The following commands need to be executed for all `VF`s: 

Example for PCI address `0000:06:06.1`:
```
echo 'vfio-pci' | sudo tee /sys/bus/pci/devices/0000:06:06.1/driver_override
echo '0000:06:06.1' | sudo tee /sys/bus/pci/drivers/vfio-pci/bind
```

## TODO: 

- [ ] @lukasfrank Check if `cloud-hypervisor-provider` and `libvirt-provider` can run on the same host

byteocean:
#680 

This enhanced init script provides extra steps to manually bind VFs to vfio-pci driver, which is automatically done by libvirt.

There are two ways to use this feature:
1) after rebooting, use this script with `--vfio-auto-bind`. It creates VFs binding to vfio-pci;
2) if you already have VFs, terminate/undefine all VMs, and run this script with `--vfio-bind-only`, it will do binding only.

### DIFF
--- a/hack/prepare.sh
+++ b/hack/prepare.sh
@@ -19,6 +19,8 @@ IS_X86_WITH_BLUEFIELD=false
 IS_ARM_WITH_BLUEFIELD=false
 IS_X86_WITH_MLX=false
 CONFIG_ONLY=false
+OPT_VFIO_AUTO_BIND=false
+VFIO_BIND_ONLY=false
 
 function log() {
 	echo "$(date +"%Y-%m-%d_%H-%M-%S-%3N") $1"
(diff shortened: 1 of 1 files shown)


hack/prepare.sh
```diff
@@ -19,6 +19,8 @@ IS_X86_WITH_BLUEFIELD=false
 IS_ARM_WITH_BLUEFIELD=false
 IS_X86_WITH_MLX=false
 CONFIG_ONLY=false
+OPT_VFIO_ASUTO_BIND=false
```
PlagueCZ, line 22: typo: `ASUTO`

hack/prepare.sh
```diff
@@ -139,6 +141,45 @@ process_multiport_eswitch_mode() {
...
+	numvfs=$(cat /sys/bus/pci/devices/$pf0/sriov_numvfs)
+	vf_offset=$(cat /sys/bus/pci/devices/$pf0/sriov_offset)
+
+	vf_in_batch_count=0
+	vf_batch_count=0
+	vf_batch_hex_index=0
+
+	modified_pci=${pf0::-3}
+
+	for ((i=$vf_offset;i<=1+$numvfs;i+=1)); do
```
PlagueCZ, line 161: This made me stop and read it again, because I usually see the traditional `for i in $(seq 1 $count)`.

hack/prepare.sh
```diff
@@ -139,6 +141,45 @@ process_multiport_eswitch_mode() {
...
+		vf_in_batch_count=$(($i%8))
+		if [ $vf_in_batch_count -eq 0 ]; then
+			vf_batch_count=$[$vf_batch_count + 1]
+			vf_batch_hex_index=$(printf '%x' $vf_batch_count)
+		fi
+		digits=${#vf_batch_hex_index}
+		if [ $digits -eq 2 ]; then
+			modified_pci=${pf0::-4}
+		fi
```
PlagueCZ, line 170: This block is related to the fact that the pci addresses are not in one block of numbers right? But I would not understand this at all without prior knowledge of the addresses, can you add some short comment atop this block about the fact? i.e. why is this complex stuff needed.

hack/prepare.sh
```diff
@@ -139,6 +141,45 @@ process_multiport_eswitch_mode() {
...
+
+		vf_pci_addr=$modified_pci$vf_batch_hex_index"."$vf_in_batch_count
+
+		log "$vf_pci_addr is the VF PCI address"
+
+		echo 'vfio-pci' | tee /sys/bus/pci/devices/$vf_pci_addr/driver_override > /dev/null
+		echo $vf_pci_addr | tee /sys/bus/pci/drivers/vfio-pci/bind > /dev/null
```
PlagueCZ, line 177: The `tee` is usually needed for `sudo tee` constructs because simple redirection will be done by the original user and not root. But you do not use sudo here, so the `tee` is doing the original intended work, i.e duplication output, which you then immediately put to `/dev/null`, so I think normal redirection is the way to go here.

hack/prepare.sh
```diff
@@ -139,6 +141,45 @@ process_multiport_eswitch_mode() {
...
+		log "Bound VF $vf_pci_addr to vfio-pci driver"
+		sleep 0.5
```
PlagueCZ, line 179: This will create one-minute delay when you use all 120+ VFs. Is that necessary?

Maybe it is, which I would believe because stuff is happening in the card, but it's surprising that it cannot handle multiple at once.

PlagueCZ (commented):
Seems the best way to implement this, just some small details I would address.

hack/prepare.sh
PlagueCZ: If the problem is `get_pattern()`, that one is only called once, so the only place where sleep should be needed is at the end of binding all VFs, not after each one though.
Os maybe increasing it to 1s and moving it out of the loop should fix the problem?

PlagueCZ (approved):
Looks great now.
Did not test on my hardware though.

guvenc (approved):

PR #696, opened 2025-07-10
Introduce DP_SUCCESS() macro

### LINKED ISSUE #695: DP_SUCCESS() macro as inverse of DP_FAILED()
The absolute majority of calls are fail-early scenarios with `DP_FAILED()` as a wrapper.

However some calls are better handled via positive test, like
```
if (!DP_FAILED(rte_hash_lookup())) {
    // do extra code when entry already present
}
```

This is visually almost identical to the inverse case dues to how small the exclamation mark is.

It would be better to also have a `DP_SUCCESS(ret)` macro to make it absolutely clear that this is the non-standard test for *positive* result.

PlagueCZ:
Added `DP_SUCCESS(ret)` macro as a replacement for `!DP_FAILED(ret)`. Cases with direct test of `ret == DP_OK` are untouched, as success is any non-negative integer, not only zero.

One case can actually now be written more concisely.

Fixes #695

### DIFF
--- a/docs/development/style_guide.md
+++ b/docs/development/style_guide.md
@@ -85,6 +85,7 @@ int caller()
   return DP_OK;
 }
 ```
+> There is also a `DP_SUCCESS(ret)` for the unlikely inverse test, e.g. `if (DP_SUCCESS(rte_hash_lookup(...)))`.
 
 ### Boolean-like calls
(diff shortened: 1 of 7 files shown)


guvenc (approved):
Thanks. The readability is better now.

Issue #697, opened 2025-07-18
NAT table synchronization between two dpservice instances


PlagueCZ:
This is an overarching issue for HA-dpservice. 

The two dpservice instances are working in active-backup mode, the second one needs to receive updates about NAT table entries. Otherwise when taking over it is unable to resume NAT communication properly.

PR #698, opened 2025-07-18
snat allocation function refactor

PlagueCZ:
For #697 I need to re-use SNAT create/delete functions. 

This PR reorganizes the code slightly for that to be easier in a followup PR.

I have separated the work into small commits so that each is easier to be reviewed as a separate change.

### DIFF
--- a/include/dp_nat.h
+++ b/include/dp_nat.h
@@ -104,7 +104,7 @@ int dp_add_neighnat_entry(uint32_t nat_ip,  uint32_t vni, uint16_t min_port, uin
 
 int dp_del_neighnat_entry(uint32_t nat_ip, uint32_t vni, uint16_t min_port, uint16_t max_port);
 
-int dp_allocate_network_snat_port(struct snat_data *snat_data, struct dp_flow *df, uint32_t vni);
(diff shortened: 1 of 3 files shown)


PlagueCZ:
Oh, the "edge case" is the fact that when previously doing a rollback (which basically never happens) `portmap_data->flow_cnt++;` was not reverted.

PlagueCZ:
I added a commit using a `goto` to a common return code because returning a positive result from the middle of the function is not common in our codebase and also any additional stuff needed to be done when returning the NAT port would need to be duplicate.

For example - changes from  #700 would require duplicating `DP_STATS_NAT_INC_USED_PORT_CNT(port)`.

Also for #697 there needs to be some `dp_send_add_nat_to_other_dpservice()` which would also then need to be done in two places.

That's why I opted for a simple `goto` to the return code. This can also be replaces by a wrapper function if preferred.

byteocean (approved):

byteocean (approved):

PlagueCZ:
I had to rebase onto #700 because otherwise the review is really misleading as you never saw the actual code that I am working on.

guvenc (approved):
Thanks

PR #700, opened 2025-07-18
Fix SNAT rollback

### LINKED ISSUE #699: Improper rollback in ipv4/nat64 snat
In functions [`dp_process_ipv4_snat()`](https://github.com/ironcore-dev/dpservice/blob/0ffd490941ecd741f647c79b803f64e9b7b0b1f9/src/nodes/snat_node.c#L73) and [`dp_process_ipv6_nat64()`](https://github.com/ironcore-dev/dpservice/blob/0ffd490941ecd741f647c79b803f64e9b7b0b1f9/src/nodes/snat_node.c#L137) there is a potential leak of NAT entry. If the flow table insertion fails, the currently created SNAT entry is not reverted and therefore will never be removed because it's the job of the flow table to age-out flows and remove SNAT entries after the right timeout.

In `dp_process_ipv6_nat64()` this is [already done for one scenario](https://github.com/ironcore-dev/dpservice/blob/0ffd490941ecd741f647c79b803f64e9b7b0b1f9/src/nodes/snat_node.c#L101), but not for the one mentioned above.

Also `DP_STATS_NAT_DEC_USED_PORT_CNT()` is already called [at the end of `dp_remove_network_snat_port()`](https://github.com/ironcore-dev/dpservice/blob/0ffd490941ecd741f647c79b803f64e9b7b0b1f9/src/dp_nat.c#L780) therefore it must not be called when doing such rollback.

---

Additional fix in snat:
Previously there was a quick-return in deletion, if portmap table had no entry, but this then jumped over the full return code that cintained for example `DP_STATS_NAT_DEC_USED_PORT_CNT()`.

PlagueCZ:
Addressed rollback problems as per issue.

Hopefully I did not misunderstand the code flow, please verify.

Fixes #699

### DIFF
--- a/src/dp_nat.c
+++ b/src/dp_nat.c
@@ -748,8 +748,10 @@ int dp_remove_network_snat_port(const struct flow_value *cntrack)
 
 	// forcefully delete, if it was never there, it's fine
 	ret = rte_hash_del_key(ipv4_netnat_portoverload_tbl, &portoverload_tbl_key);
-	if (DP_FAILED(ret) && ret != -ENOENT)
+	if (DP_FAILED(ret) && ret != -ENOENT) {
+		DPS_LOG_ERR("Cannot delete portoverload key", DP_LOG_RET(ret));
 		return ret;
+	}
 
 	dp_copy_ipaddr(&portmap_key.src_ip, &flow_key_org->l3_src);
 	portmap_key.iface_src_port = flow_key_org->src.port_src;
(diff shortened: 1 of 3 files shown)


PlagueCZ:
The second commit is an addition as it occurred to me that having the `INC` macro used in `snat_node.c` while the `DEC` macro used in `dp_nat.c` is asking for trouble, so I moved them to be in one place (inside NAT functions themselves).

byteocean (approved):
Approved as communicated

src/dp_nat.c
```diff
@@ -759,18 +763,22 @@ int dp_remove_network_snat_port(const struct flow_value *cntrack)
...
+				return DP_ERROR;
+			}
+			rte_free(portmap_data);
 		}
-		rte_free(portmap_data);
+	} else {
+		DPS_LOG_ERR("Cannot lookup portmap key", DP_LOG_RET(ret));
+		if (ret != -ENOENT)
+			return ret;
+		// otherwise already deleted, finish
```
guvenc, line 781: I understand that this control flow is intentional now so that it can hit the INC macro at the end but this would also cause the new code to hit the lines for `ret == ENOENT` case.

> 	created_port = dp_get_port_by_id(cntrack->created_port_id);
	if (!created_port)
		return DP_ERROR;

Was this intentional ?

PlagueCZ: Yes. This is the commit that "fixes SNAT edge case" because previously it could have found a portmap entry (and immediately deleted it), but not found portoverload entry and then return without actually decreasing the  counter.

guvenc (approved):

PR #709, opened 2025-08-13
Use underlay prefix and two harcoded bytes for traffic isolation

### LINKED ISSUE #711: Adjust isolation rules to eliminate ipv6 encapsulation dependency
Isolation rules of the dpservice should not check the fact that whether the traffic is encapsulated, it should rather check a predefined destination ipv6 range for the traffic it wants to manage.

Motivation:
- There can be other ipv6 encapsulaters on the node where dpservice is running.
- This can simplify several isolation rules we install. (Standard rule, virtual service rule)

guvenc:
Do not isolate only the ip in ipv6 encapsulated packets but also the specific subnet used by dpservice on a given node.


PlagueCZ (changes requested):
Small details found.

Also I would like to discuss the prefix idea

PlagueCZ:
I finished the needed changes, moved to range `d000`..`dfff`, ported changes to async flow rules and then removed all rules for virtual services as they are no longer necessary.

Code is much simpler everywhere now.

I documented my prefix schema in `docs/deployment/` and added support for `vnf_type` to be encoded in the address (optional in meson). I reordered `vnf_type` because I implemented this change in metalnet already and ordered it to be easily remembered, only later finding that the same enum already exists in dpservice. This is also why the one test needed to change (wcmp hash changed).

I tested on pytest with HW (mellanox) tests, and also deployed in OSC, both normal and multiport setup, i.e. both sync and async rules are fully tested and working fine. I believe Florin will be testing the running of a separate IPIP tunnel in parallel to dpservice on Monday.

I also moved `dp_rte_flow_init` module to `dp_rte_flow_isolation` to be consistent with the async version.

guvenc:
LGTM

PlagueCZ (approved):

PR #710, opened 2025-08-14
prepare script improvements

### LINKED ISSUE #712: Improve prepare.sh script to support Bluefield
Improve prepare.sh script to support Bluefield and the script should be more robust when parsing the command line output.

guvenc:
prepare script improvements

### DIFF
--- a/hack/prepare.sh
+++ b/hack/prepare.sh
@@ -11,7 +11,7 @@ set -Eeuo pipefail
 
 OPT_MULTIPORT=false
 
-BLUEFIELD_IDENTIFIERS=("MT_0000000543", "MT_0000000541")
+BLUEFIELD_IDENTIFIERS=("MT_0000000543" "MT_0000000541")
 MAX_NUMVFS_POSSIBLE=126
 NUMVFS_DESIRED=126
 CONFIG="/tmp/dp_service.conf"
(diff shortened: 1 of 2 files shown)


hack/prepare.sh
```diff
@@ -158,8 +158,6 @@ function vfio_bind() {
 
 	modified_pci=${pf0::-3}
 
-	# pci address of vfs are organized in a block of 8
-	# e.g., 3b:00.0 ... 3b:00.7, 3b:01.0 ... 3b:01.7, etc.
```
PlagueCZ, line 162: This comment was here to explain the strange for loop, because for those that do not know the VF numbering scheme it makes no sense just by reading the loop.

hack/prepare.sh
```diff
@@ -345,5 +381,9 @@ validate
 get_pfs
 detect_card_and_arch_type
 validate_pf
+# Bluefield relies on host side creating VFs, so wait here for systemd to settle and dont rush with the next steps
+if [[ "$IS_ARM_WITH_BLUEFIELD" == "true" ]]; then
+	udevadm settle
+fi
 create_vf
-make_config
+make_config
```
PlagueCZ, line 389: Please do not remove endlines, when I sometimes `cat` files it messes up my terminal prompt :)

PlagueCZ (commented):
Hard for me to test give then state of my work and the fact that we use ignition mostly, but it looks OK from what it's doing.

hack/prepare.sh
guvenc: Makes totally sense. I will add all the removed original comments.

PlagueCZ (approved):
Changes make sense, cannot test in OSC as we use ignition for this part.

PR #714, opened 2025-08-27
Feature/vf mac change func

### LINKED ISSUE #713: Unify MAC address handling in ARP/DHCP/ND
Currently, ARP only ever sets the address once and then the value is reused for the lifetime of the VF.

In DHCP and ND, the address is always overwritten by the value taken from packets.

Neither approach is inherently bad, but it is inconsistent.

PlagueCZ:
I created a wrapper function to set the VF MAC address. This will later be helpful for synchronizing the MAC address with the backup dpservice.

MAC address is only changed once per-lifetime of the VF. The old value is kept and reused as a "suggestion" until ARP/DHCP/ND arrives.

The initial value is set in the process init stage to the MAC address of the representor.

I also renamed `dp_set_neigh_mac()` as it was meant as the "neighboring router mac", i.e. neighbor of a PF, never a VF.

Fixes #713

### DIFF
--- a/include/dp_iface.h
+++ b/include/dp_iface.h
@@ -4,6 +4,8 @@
 #ifndef __INCLUDE_DP_IFACE_H__
 #define __INCLUDE_DP_IFACE_H__
 
+#include <rte_ether.h>
+#include <stdint.h>
 #include "dp_port.h"
 
 #ifdef __cplusplus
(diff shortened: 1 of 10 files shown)


guvenc (approved):
LGTM

guvenc:
@PlagueCZ  could you please rebase to main ?

PR #716, opened 2025-09-02
Fix default route detection

### LINKED ISSUE #715: NAT default route detection edge-case
NAT only translates packets that are going through the default route, based on DP_FLOW_SOUTH_NORTH as seen in [snat_node.c:177](https://github.com/ironcore-dev/dpservice/blob/786ca5ea03348909696a82a4399699891dec6820/src/nodes/snat_node.c#L177).

This is set in [ipv4_lookup_node.c:56](https://github.com/ironcore-dev/dpservice/blob/786ca5ea03348909696a82a4399699891dec6820/src/nodes/ipv4_lookup_node.c#L56).

Basically, if the destination route IP is 0.0.0.0 then NAT is used.

This is working for most cases except, when we introduce a route that matches this IP, e.g. `0.0.0.0/1`.

By definition, this route has precedence over the default route, because the length/depth is bigger, however the code does not use depth match, only IP match.

There is a pytest prepared in a [branch commit](https://github.com/ironcore-dev/dpservice/commit/4301e8df7ee2f942b9629dc5112e8e4c76652b79).

---
Yes, this basically prevents NATting and sends all traffic to some catch-all address, which is something OSC needed in a special case. Ignoring that however, route matching should by definition use length, so this should be possible (even though it is strange).

### LINKED ISSUE #718: Pytest suite  expect_error() bug
Since 92981a3f6155d1a36b6862caa64cf16ba408a6d1 the test suite does not correctly fail when expecting gRPC error an no error comes.

PlagueCZ:
Default route detection is now based on route depth/length not IP itself.

The same problem was technically present in IPv6, but there we only do NAT for NAT64 addresses which prevents the bug from manifesting itself.

I was unable to think up a good test for it because of the NAT64 restrictions, but I tested via debug logs inside ipv6_lookup_node.c

Fixes #715
Fixes #718


byteocean (approved):
LGTM now as communicated.

test/local/test_nat.py
```diff
@@ -1,12 +1,51 @@
...
+		threading.Thread(target=vm_sender, args=(str(net.broadcast_address+2),)).start()
+		pkt = sniff_packet(PF0.tap, is_udp_pkt)
+		assert pkt[IP].src != VM1.ip, \
+			"IP not translated"
+
+	grpc_client.delroute(vni1, route)
+
+def test_nat_default_route(prepare_ifaces, grpc_client):
+	nat_ul_ipv6 = grpc_client.addnat(VM1.name, nat_vip, nat_local_min_port, nat_local_max_port)
+	try_ip_range(grpc_client, "128.0.0.0/1")
```
guvenc, line 44: Nice idea to test with X.0.0.0/1 for checking the default route behaviour with /0.

guvenc (approved):
LGTM

PR #719, opened 2025-09-03
Switch from /tmp to /run to DP_CONF default path

### LINKED ISSUE #717: Config written to /tmp gets cleared after 10 days
**Describe the bug**

dpservice writes a config file to tmp (`/tmp/dp_service.conf`) which should be cleared on reboots. Recently the configuration for tmp changed and files will now be cleared after being untouched for ten days causing dpservice to fail.

**To Reproduce**

On a recent debian(-based) system start dpservice, wait 10 days and see the issue.

**Expected behavior**

dpservice does not write important config data to directories which are occasionally cleared while the system is running.

According to the [Filesystem Hierarchy Standard 3.15](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s15.html) the proper location for such files is `/run`.

**Screenshots**
 n/a

**Additional context**

I'm already preparing a fix which moves the config file to `/run`.

maxmoehl:
On modern debian-based systems files may be cleared from /tmp after some
time. The recommended location for files which should be kept as long as
the system is running is the /run directory.

This commit changes the relevant references to load the config file from
/run/dpservice/dpservice.conf by default instead of /tmp/dp_service.conf.
The old behavior can be restored by setting DP_CONF to
/tmp/dp_service.conf.

The hack/prepare.sh was updated to respect the DP_CONF environemnt
variable as well.

Resolves: #717


hack/prepare.sh
```diff
@@ -330,6 +330,7 @@ function make_config() {
 	conf_vf_pattern="$(get_pattern "${devs[0]}")"
 	conf_ipv6="$(get_ipv6)"
 
+	mkdir -p "$(dirname "$CONFIG")"
```
PlagueCZ, line 333: While I totally understand the need for creating the directory somewhere, and the fact this is the simplest way to do it, I am wary of situations where DP_CONF is used. Then this script would call `mkdir -p $(dirname )` on an arbitrary quoted string and do it as root.

Not saying it's a big problem, but I would like to mention it, because it make me feel uneasy at first glance, may be just paranoia.

PlagueCZ (changes requested):
I am unable to currently test this in OSC, but it seems that I can just define DP_CONF and make it compatible with current production setup, so it looks fine.

You are missing changes to `config/default/dp-service.yaml` where `/tmp` is still mounted but `/run/dpservice` is not. This will also remove the need for doing a `mkdir` inside the script, because the directoyr will be craeted by kubernetes.

hack/prepare.sh
maxmoehl: That's definitely a valid point. The alternative I see is something like:

```
if ! test -d "$(dirname "$CONFIG")"; then
    echo "error: config directory does not exist"
    exit 1
fi
```

But for automated deployments you would then need some custom script to setup the directory first since `/run` is wiped on reboots.

guvenc: The change(s) in `config/default/dp-service.yaml` is missing. If you specify the hostPath in deployment yaml of the dpservice with type `DirectoryOrCreate` then the kubelet would create the directory, if it doesnt exist and you can eliminate the "mkdir" in the script which tries to create the directory with arbitrary given string.

https://github.com/ironcore-dev/dpservice/blob/786ca5ea03348909696a82a4399699891dec6820/config/default/dp-service.yaml#L103

guvenc: @maxmoehl  in case you missed the last comment.

PlagueCZ (commented):
Looks great now.

I should test in OSC, but I would need to prepare three overlay YAMLs we have (we have three separate configs) and the environment is somewhat flaky at this point, and I am short on time, so let's trust how it looks.

guvenc (approved):
Thanks. LGTM

PR #722, opened 2025-09-12
Support VF TX rate limiting using netlink

### LINKED ISSUE #679: TX rate limiting on VFs based on byte counts
Performing TX rate limiting on VFs via sysfs parameters was not valid anymore #488. A suggested alternative approach by Nvidia is to use the ip command `ip link set ens5f0np0 vf 0 max_tx_rate 10000 min_tx_rate 10000` to configure VF's max tx rate.

- [x] verify if the suggested command works or not
- [ ] integrate the suggested approach into dpservice

byteocean:
Address #679 

How to test:
1) start dpservice
2) run commands:
```
./dpservice-cli create interface --id=vm3 --vni=66 --ipv4=172.32.4.9 --ipv6=2003::123 --device=0000:3b:00.0_representor_vf0 --total-meter-rate=5000 --public-meter-rate=500

ip link show enp59s0f0np0
6: enp59s0f0np0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP mode DEFAULT group default qlen 1000
    link/ether 04:3f:72:e8:cf:ca brd ff:ff:ff:ff:ff:ff
    vf 0     link/ether 00:00:00:00:00:00 brd ff:ff:ff:ff:ff:ff, tx rate 5000 (Mbps), max_tx_rate 5000Mbps, spoof checking off, link-state auto, trust off, query_rss off
    vf 1     link/ether 00:00:00:00:00:00 brd ff:ff:ff:ff:ff:ff, spoof checking off, link-state auto, trust off, query_rss off
    
 
./dpservice-cli delete interface --id=vm3

ip link show enp59s0f0np0
6: enp59s0f0np0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP mode DEFAULT group default qlen 1000
    link/ether 04:3f:72:e8:cf:ca brd ff:ff:ff:ff:ff:ff
    vf 0     link/ether 00:00:00:00:00:00 brd ff:ff:ff:ff:ff:ff, spoof checking off, link-state auto, trust off, query_rss off
    vf 1     link/ether 00:00:00:00:00:00 brd ff:ff:ff:ff:ff:ff, spoof checking off, link-state auto, trust off, query_rss off
```

killing dpservice / crashing leaves the configured rate untouched until new dpservice configures it when creating interface.

### DIFF
--- a/src/dp_netlink.c
+++ b/src/dp_netlink.c
@@ -15,7 +15,7 @@
 #include "dp_util.h"
 #include "dp_port.h"
 
-static int dp_read_neigh(struct nlmsghdr *nh, __u32 nll, struct rte_ether_addr *neigh,
+static int dp_nl_read_neigh(struct nlmsghdr *nh, __u32 nll, struct rte_ether_addr *neigh,
 						 const struct rte_ether_addr *own_mac)
 {
 	struct rtattr *rt_attr;
@@ -44,7 +44,7 @@ static int dp_read_neigh(struct nlmsghdr *nh, __u32 nll, struct rte_ether_addr *
 	return DP_ERROR;
 }
 
-static int dp_recv_nl_neigh_msg(struct sockaddr_nl sock_addr, int sock, char *buf, int bufsize)
+static int dp_nl_recv_neigh_msg(struct sockaddr_nl sock_addr, int sock, char *buf, int bufsize)
 {
 	struct nlmsghdr *nh;
 	ssize_t recv_len;
@@ -117,13 +117,13 @@ int dp_nl_get_pf_neigh_mac(uint32_t if_idx, struct rte_ether_addr *neigh, const 
 		goto cleanup;
 	}
 
-	reply_len = dp_recv_nl_neigh_msg(sa, sock, reply, sizeof(reply));
+	reply_len = dp_nl_recv_neigh_msg(sa, sock, reply, sizeof(reply));
 	if (reply_len < 0) {
 		DPS_LOG_ERR("Cannot receive message from netlink", DP_LOG_NETLINK(strerror(reply_len)));
 		goto cleanup;
 	}
 
-	ret = dp_read_neigh((struct nlmsghdr *)reply, reply_len, neigh, own_mac);
+	ret = dp_nl_read_neigh((struct nlmsghdr *)reply, reply_len, neigh, own_mac);
 
 cleanup:
 	close(sock);
--- a/src/dp_netlink.c
+++ b/src/dp_netlink.c
@@ -44,7 +44,7 @@ static int dp_read_neigh(struct nlmsghdr *nh, __u32 nll, struct rte_ether_addr *
 	return DP_ERROR;
 }
 
-static int dp_recv_msg(struct sockaddr_nl sock_addr, int sock, char *buf, int bufsize)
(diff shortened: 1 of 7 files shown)


PlagueCZ (changes requested):
Code seems quite good, just a few suggestions and a potential problem.

I cannot find documentation for rate limiting (grepped docs/), can you write a short one?
I forget what is the difference between total and public.

I tested on my Mellanox locally with my 2VMs and was successful at limiting `iperf` using total rate. Public rate did nothing for me.

There is no pytest for this feature. Which of course would be hard to test the rate itself, but can we at least test the output of `ip link` maybe? I'll try to cook up something in a branch to incorporate into `gRPC` test file.

I would also like to test in OSC? but it takes time, so I would do it at the end if that is ok.

src/dp_netlink.c
```diff
@@ -129,3 +129,95 @@ int dp_get_pf_neigh_mac(uint32_t if_idx, struct rte_ether_addr *neigh, const str
 	close(sock);
 	return ret;
 }
+
+
+static int dp_nl_send_recv_vf_rate_msg(int sock, struct dp_nl_vf_rate_req *req)
+{
+	// ssize_t  sent_len;
```
PlagueCZ, line 136: some leftover

src/dp_netlink.c
```diff
@@ -129,3 +129,95 @@ int dp_get_pf_neigh_mac(uint32_t if_idx, struct rte_ether_addr *neigh, const str
...
+		return DP_ERROR;
+	}
+
+	sent_len = send(sock, req, req->nl.nlmsg_len, 0);
+	if (sent_len < 0 || sent_len != req->nl.nlmsg_len) {
+		DPS_LOG_ERR("Failed to send VF rate message", DP_LOG_NETLINK(strerror(errno)));
+		return DP_ERROR;
+	}
+
+	// Wait for ACK, ensure it gets at least NLMSG_LENGTH(0) bytes
```
PlagueCZ, line 157: This comment is lying, we are testing different length.

src/dp_netlink.c
```diff
@@ -129,3 +129,95 @@ int dp_get_pf_neigh_mac(uint32_t if_idx, struct rte_ether_addr *neigh, const str
...
+	reply_len = recv(sock, reply, sizeof(reply), 0);
+	if (reply_len < (ssize_t)NLMSG_LENGTH(sizeof(struct nlmsgerr))) {
+		DPS_LOG_ERR("Cannot receive ACK from netlink", DP_LOG_NETLINK(strerror(errno)));
+		return DP_ERROR;
+	}
+
+	nh = (struct nlmsghdr *)reply;
+	if (nh->nlmsg_type == NLMSG_ERROR) {
```
PlagueCZ, line 165: After I read thought the man page, this should actually always be this type (guaranteed by requesting ACK), so maybe return error if not the case?

PlagueCZ (changes requested):

PlagueCZ (approved):
Looks great, documentation now mentions Tx only so curl not being limited shold stop being surprising to me :)

guvenc (changes requested):
Actually in very good state. Couple of nit picks and a suggestion to eliminate the string operations. Thanks !

guvenc (approved):

PR #723, opened 2025-09-18
Reduce the size of the string used to name internal memory RIB tables

### LINKED ISSUE #725: Size of the string used to name internal memory RIB tables may break on Bluefield
Bluefield systems do not have NUMA so that the index used for NUMA becomes "-1". As this index is 
part of the RIB table naming string, this can trigger an error in mlx5 PMD memory allocation routines.
As this one extra character can be over a hidden 25 character limit of PMD driver.

guvenc:
The allocated memory table name used by the RIB structure has a hidden limit of 25 characters in PMD. This becomes a problem when 8 digit VNIs are used and there's no NUMA(like on Bluefields) and socket_id becomes -1 and uses two characters.
So reduce the rest of the string to have enough room even with -1 in the string.


PlagueCZ (approved):
Short and simple, looks good.

PlagueCZ (approved):

PR #724, opened 2025-09-22
Remove unused tx-rate defines

byteocean:
As the sysfs approach is not used anymore, these defines should be also removed.


PlagueCZ (approved):
Compiles, so LGTM

guvenc (approved):

PR #731, opened 2025-10-20
Fix telemetry free null-pointer bug

### LINKED ISSUE #730: Crash in `rte_graph_cluster_stats_destroy()`
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fba31d93047 in rte_graph_cluster_stats_destroy (stat=0x0) at ../lib/graph/graph_stats.c:448
```

Reason is simple, as opposed to most other rte_free style calls, this is not safe to call on a NULL pointer, as clear from the source code in `lib/graph/graph_stats.c`

PlagueCZ:
Simply added a check that I previously thought is included in the RTE call, as all similar calls do check that.

Fixes #730

### DIFF
--- a/src/dp_telemetry.c
+++ b/src/dp_telemetry.c
@@ -26,7 +26,7 @@ struct dp_telemetry_htable {
 static struct dp_telemetry_htable *tel_htables = NULL;
 static int tel_htable_count = 0;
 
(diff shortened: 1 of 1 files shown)


byteocean (approved):
LGTM.

guvenc (approved):
Thanks

Issue #329, opened 2023-07-12
Zero Downtime Updatability for dp-service


guvenc:
There is the need for a concept of "Zero downtime updatability" for dp-service. At the moment update of dp-service would cause all the running connections break and re-establish. This may not be desirable in many customer use-cases, although the down-time would be less than a minute with the current dp-service implementation.

Other than developing a concept in dp-service and implementing it, the following ideas could be a workaround till a final solution is available:
- A concept of VM immigration where all VMs can be immigrated to another node without traffic disruption and dp-service can be updated on a VM-free node.
- If the workloads in the VM are running using Kubernetes then the workload pods can be "drained" to other nodes and dp-service can be updated on the node where the pods are completely drained.

PR #735, opened 2025-11-04
Implement dpservice-ha synchronization

### LINKED ISSUE #697: NAT table synchronization between two dpservice instances

PlagueCZ:
This PR implements a mechanism for synchonizing the active and standby instances of dpservice. 
 - synchronizes current NAT connections
 - synchronizes current virtual service connections
 - synchronizes VF neightbor MACs
 - uses TAP devices in a bridge for synchronization
 - connection tracking is not synchronized, but re-created when needed
 
Pytest suite and documentation are part of this PR

Fixes #697

Connected to #329


PlagueCZ:
Solved. For some reason in a container pytest is not automatically refreshing interfaces. And when the test re-starts secondary dpservice, which uses the same interface names all the time, its interfaces are not properly registered in pytest.

guvenc (changes requested):
I had the chance to test it regarding the backward compatibility and the PR looks good to me apart from the small typo.
Thanks !

guvenc (approved):

Issue #736, opened 2025-11-10
Improve overlay observability and infrastructure operations quality of life


PlagueCZ:
This is an overarching issue for small changes, like:
 - improved logging in certain situations (failure states)
 - improved `dpservice-*` tools
 - etc.

PR #737, opened 2025-11-10
Add port range to NAT full warning

PlagueCZ:
The NAT being full warning is sometimes caught by Loki or when investigating a production problem. 

It would be helpful to immediately see how big the range actually is, because it is not uncommon that the user/customer sets the range too small (by having unnecessarily large subnet).

This information is visible elsewhere, but having it here is an immediate help in certain situations.

Issue #736


byteocean (approved):

guvenc:
@PlagueCZ  Please rebase to main.

guvenc (approved):
LGTM

PR #739, opened 2025-11-10
Remove unneeded packet type specifications in pytest

### LINKED ISSUE #738: Pytest specifies packet types manually
Scapy can recognize the needed packet type (or next-header field) automatically from the connected packet, e.g.:
```
pkt = Ether() / IP()
```
will correctly get ethertype IP.

Removing manual setting of these fields make the test more readable, reduces hardcoded integer values and makes reusing the code easier as we do not need to think about the additional values to fill.

PlagueCZ:
Simply removed unneeded arguments from packet creation where possible.

Fixes #738

### DIFF
--- a/test/local/helpers.py
+++ b/test/local/helpers.py
@@ -212,8 +212,8 @@ def stop_process(process):
 
 
 def _send_external_icmp_echo(dst_ipv4, ul_ipv6):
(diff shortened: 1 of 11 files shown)


guvenc (approved):
LGTM

PR #740, opened 2025-11-10
Log which loadbalancer already exists when adding fails

PlagueCZ:
If adding a loadbalancer fails with ALREADY_EXISTS, it would be nice to know what is the ID of the existing LB, makes it easier to investigate.

Of course this information already is in some objects, but since we are already doing a hashtable lookup, this is basically free information and it does make operations easier.

Issue #736


guvenc (approved):
LGTM

PR #741, opened 2025-11-10
Add tcpdump_helpers.inc to dpservice container

PlagueCZ:
When using `dpservice-dump`, the pcap-style filter can either specify an overlay IPv4 or underlay IPv6.

This means when tracing a packet over the graph, the user must manually switch from underlay to overlay, thus losing the full trace.

Including these macros (already used on every node in OSC for tcpdump), the user can do
```
dpservice-dump --nodes "" --filter "host 1.2.3.4 or $(ipip_host 1.2.3.4)"
```
and see the whole packet trace over the graph from start to finish.

Issue #736

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -179,7 +179,8 @@ COPY --from=builder \
 COPY --from=builder /usr/local/lib /usr/local/lib
 RUN ldconfig
 
-# Ensure bash-completion is working in operations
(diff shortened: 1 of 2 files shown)


guvenc (approved):
LGTM

guvenc:
@PlagueCZ 
Could you please rebase to main ?

PR #743, opened 2025-11-10
Remove unneded mount of hugepages in init container

### LINKED ISSUE #721: prepare script mounting hugepages
Looking at current `prepare.sh` it does not use hugepages, yet they are still being mounted. I think it's some obsolete mount from previous version of the script.

I tested in osc after removing that mount, but we are preparing the nodes differently so if someone could confirm the fact it can be removed, that would be good.

PlagueCZ:
Removes `/dev/hugepages` mount into init container as it no longer uses it.

Tested in an OSC cluster, though there I know where the hugepages are being set-up.

Fixes #721

### DIFF
--- a/config/default/dpservice.yaml
+++ b/config/default/dpservice.yaml
@@ -25,8 +25,6 @@ spec:
         command:
         - prepare.sh
         volumeMounts:
-        - mountPath: /dev
-          name: hugepages
         - mountPath: /run/dpservice
           name: run
         securityContext:


damyan:
Do I get it right, that if we remove this, then some preprovisioning must be in-place to mount the hugepages? (we *do* need hugepage mount). Preprovisiong meaning either buildiing in the functionality in Gardenlinux or in some agent like `gardener-node-agent`.
/cc @guvenc @byteocean

guvenc (approved):
LGTM

PR #744, opened 2025-11-11
Add option to only dump a number of packets using dpservice-dump

PlagueCZ:
Just like `tcpdump -c`, I introduced `dpservice-dump -c` to limit the number of dumped packets. 

This is extremely helpful when tracing a stream of packets under high traffic, where one wants to simply see what is happening on a given IP for example.

Issue #736


guvenc (approved):
LGTM

PR #746, opened 2025-11-15
List all routes and lbtargets using gRPC

### LINKED ISSUE #745: Add ability to list also all routes and lbtargets
Previously, the ability to list all NAT entries and all Prefixes (i.e. without specifying the NAT ip or interface) has bee introduced.

It is helpful for quick comparison/check of dpservice against metalnet for example.

Now with HA, there can be a bug in orchestration, where `dpservice-a` and `dpservice-b` have different state of orchestration.

While it is certainly possible to iterate over all loadbalancers and interfaces and then call a list of targets/prefixes on each one, having one call to do this would be really helpful, because doing calls to a running dpservice by human operators during production is inherently risky (and when done on multiple nodes even slow).

PlagueCZ:
I added the ability to list all loadbalancer targets and all prefixes (incl. lbprefixes), in the same fashion as previously has been done with NATs, i.e. by specifying an "empty" identifier.

For prefixes, the drawback is that in the response, the VNI is not valid, it is the "empty" one.
 - given the fact that target VNI is always the same as VNI anyway, this is a non-issue
 
For LB targets, the actual LB id is missing from the output.
 - not ideal, but for checking whether both dpservices are orchestrated the same, this is enough.

Both of the above is easily fixable, but requires changes to the gRPC protocol, which I would rather not do. And given the fact that this already covers the intended purpose (quick check of HA dpservices), the need for the change is low in my opinion.

Fixes #745

### DIFF
--- a/cli/dpservice-cli/cmd/list_loadbalancer_targets.go
+++ b/cli/dpservice-cli/cmd/list_loadbalancer_targets.go
@@ -54,11 +54,6 @@ func (o *ListLoadBalancerTargetOptions) AddFlags(fs *pflag.FlagSet) {
 }
 
 func (o *ListLoadBalancerTargetOptions) MarkRequiredFlags(cmd *cobra.Command) error {
-	for _, name := range []string{"lb-id"} {
-		if err := cmd.MarkFlagRequired(name); err != nil {
-			return err
-		}
-	}
 	return nil
 }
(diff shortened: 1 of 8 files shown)


guvenc (approved):
LGTM. Thanks for the tests as well.

PR #749, opened 2025-12-04
Increase TX_QUEUE_SIZE

### LINKED ISSUE #748: Unable to send packets under heavy load
When dpservice is under high load, more precisely the packet-per-second rate gets high (~200kp/s) the following errors start appearing:
```
Not all packets transmitted successfully, node: tx-1
```

For `tx-2` and onwards, this can be caused by a VM not being fast enough in reading packets out. But `tx-1` is actually PF1 which would mean the switch is unable to pull packets out, which is frankly ridiculous. In our setup we have `--wcmp 0` that's why we do not ever see this for `tx-0`.

This was not an issue previously because we were using un-optimized image where dpservice was the slowest link in the chain.

Alsi this change was not present:
https://github.com/ironcore-dev/dpservice/commit/965d181244d98778fa088b8a5f65b57c067579b0

PlagueCZ:
Testing this is a bit complicated, because VM is the weakest link usually, so it requires many many VMs, or alternatively, one can send the packets out 10 times. This of course breaks communication, however in local testing works as intended. 

I was able to reach 200-300 kp/s and then the error messages start appearing.

Increasing the size of TX queue by a factor of 4 makes my local tests start failing at 1 Mp/s, and in our real-life scenario makes all problems go away, only the occasional VM unable to cope with the traffic.

In real-life deployment, factor of 4 was still occasionally producing an error, factor of 8 did not. The memory allocation is not that big, so I would recommend using the larger factor for future-proofing. (130-200 MB increase in `alloc_size`, but still fits into 1GiB total heap size, OSC reserves 2GiB for dpservice anyway).

Increasing more is not really possible because we run into some DPDK limit, fortunately this seems to be enough.
```
EAL log message, eal_msg: mlx5_net: port 0 Tx WQEBB count (65536) exceeds the limit (32768), try smaller queue size
```

Fixes #748

### DIFF
--- a/include/dpdk_layer.h
+++ b/include/dpdk_layer.h
@@ -37,8 +37,9 @@ extern "C" {
 // Seen recommendations to keep this at 2*RTE_GRAPH_BURST_SIZE or 4*RTE_GRAPH_BURST_SIZE
 #define DP_RX_QUEUE_SIZE (4 * RTE_GRAPH_BURST_SIZE)
 // Seen this recommended to be bigger than Rx because multiple Rx streams share the same Tx
-// but in our case we are only using one worker thread, so there is no concurrent Tx
-#define DP_TX_QUEUE_SIZE DP_RX_QUEUE_SIZE
+// in our case we are only using one worker thread, so the original thought was that there is no concurrent Tx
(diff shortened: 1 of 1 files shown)


guvenc (approved):
LGTM

PR #752, opened 2025-12-19
Optionally allow ULA addresses in prepare.sh

### LINKED ISSUE #751: prepare.sh fails in air-gapped environment
In our air-gapped setup (and also virtual lab) we are using ULA addresses on `lo`.

This will cause an error in IPv6 resolving code inside `prepare.sh`.

It would be nice to also allow ULA.

PlagueCZ:
This PR adds an `--allow-ula` option to `prepare.sh` to make it also accept ULA IPv6 on `lo`.

Fixes #751

### DIFF
--- a/hack/prepare.sh
+++ b/hack/prepare.sh
@@ -11,6 +11,7 @@ set -Eeuo pipefail
 
 OPT_MULTIPORT=false
 OPT_SYNC_BRIDGE=false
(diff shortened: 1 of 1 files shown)


byteocean (approved):
clean solution!

guvenc (approved):

PR #753, opened 2025-12-23
Fix reuse issue

byteocean:
address #733

I assume, one of failure reasons is that, `dp_conf_generate.py` under `/hack` shall not contain license info, as this path is configured to be ignored.


guvenc (approved):
LGTM

Issue #756, opened 2026-02-23
prepare script needs to take care of extra interfaces on supermicro servers


byteocean:
In addition to the two PF ports of Bluefield card installed onto super micro servers, there are two extra interfaces of Nics, which are currently not properly excluded in the prepare script. Thus it leads to the faulty config file.

```
no-stats
pf0 ens4047f0np0
ens1025f0np0
pf1 ens4047f1np1
ens1025f1np1
vf-pattern ens1025f0r


devlink -js dev | jq -r '.dev | keys[] | select(startswith("pci/")) | sub("pci/"; "")'
0000:37:00.0
0000:37:00.1
0000:90:00.0
0000:90:00.1
```

byteocean:
resolved by #757

PR #757, opened 2026-03-04
Adapt prepare script for a machine with two Nics

byteocean:
address #756. 

More generally speaking, the script works for a machine equipped with one mellanox card and one non-mellanox card.

### DIFF
--- a/hack/prepare.sh
+++ b/hack/prepare.sh
@@ -38,17 +38,30 @@ function get_pfs() {
 }
 
 function detect_card_and_arch_type() {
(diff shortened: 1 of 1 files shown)


PlagueCZ (approved):
I tested on my local machine and on two of our clusters, the resulting config has empty diff.

We however now completely disable the other mellanox card, so I cannot test that state anymore (not easily anyway).

Given the fact that this change would not be able to address that situation, I would consider this PR tested and compatible.

PR #769, opened 2026-03-19
Accept all MAC changes from the VM

### LINKED ISSUE #768: No VM traffic after host restart
When a hypervisor node is restarted and dpservice comes back up, no VM traffic is working.

Reason being the reply comes to the VM but gets dropped. This is because the MAC address is wrong.

PlagueCZ:
Fixes #768

In a previous PR #714 I changed the behavior of MAC address discovery for the VM.

The idea was to only update the MAC once, so prevent unnecessary memory copy.

I knew that the user can change a MAC manually in the VM, but that was really not a concern.

What I missed though is the fact, that iPXE and systemd both use different MAC address:
1. There is the MAC address of the VF (can be seen on host)
2. iPXE starts the VM and requests DHCP - with the VF's original MAC
3. iPXE downloads kernel, initramfs and squashfs and starts the image
4. systemd starts DHCP but due to `MACAddressPolicy=persistent` in `/usr/lib/systemd/network/99-default.link` it generates a new MAC
5. dpservice refuses to update it because of PR #714
6. kernel drops all incoming traffic due to MAC mismatch

I was able to make this happen on many host machine reboots. After applying this PR I am unable to do it

### DIFF
--- a/src/dp_port.c
+++ b/src/dp_port.c
@@ -758,9 +758,11 @@ int dp_port_meter_config(struct dp_port *port, uint64_t total_flow_rate_cap, uin
 
 void dp_l2_addr_set(struct dp_port *port, const struct rte_ether_addr *l2_addr)
 {
-	rte_ether_addr_copy(l2_addr, &port->neigh_mac);
+	if (!rte_is_same_ether_addr(l2_addr, &port->neigh_mac)) {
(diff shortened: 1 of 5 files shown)


byteocean (approved):
Thanks for explaining. LGTM.

PR #772, opened 2026-03-31
Introduce node-ip for prepare.sh script

byteocean:
Add one extra starting flag to specify the node IP used to parse for the underlaying address prefix. 

background: In some cases, loopback interface has two IPv6 addresses. The current prepare script always takes the first one for parsing. It helps in deployment to overwrite this behaviour by specifying the node IPv6 address.


hack/prepare.sh
```diff
@@ -438,6 +443,10 @@ while [[ $# -gt 0 ]]; do
 		--allow-ula)
 			OPT_ALLOW_ULA=true
 			;;
+		--node-ip)
+			shift
```
PlagueCZ, line 443: Not a bash getopt expert, but why is there only this shift and not for all of the arguments?

hack/prepare.sh
```diff
@@ -361,7 +362,11 @@ function make_config() {
 	conf_pf0="$(get_ifname 0)"
 	conf_pf1="$(get_ifname 1)"
 	conf_vf_pattern="$(get_pattern "${devs[0]}")"
-	conf_ipv6="$(get_ipv6)"
+	if [[ -n "$OPT_NODE_IP" ]]; then
+		conf_ipv6="$OPT_NODE_IP"
+	else
+		conf_ipv6="$(get_ipv6)"
+	fi
```
PlagueCZ: This is a good candidate for [bash parameter expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) `${parameter:−word}`

Untested, but this should be it:
```
conf_ipv6=${OPT_NODE_IP:-$(get_ipv6)}
```

PlagueCZ (changes requested):

hack/prepare.sh
byteocean: thanks for the tip. it is used now.

hack/prepare.sh
byteocean: because other arguments do not take value input, but this one does. thus, it needs to take an 'inner shift' to move to the input value.

PlagueCZ (approved):

PR #774, opened 2026-03-31
Add DPSERVICE_VERSION build arg to override version during Docker build

### LINKED ISSUE #773: Allow dpservice version string to be customizable
In OSC, component version also includes the overall OSC package version, like `v1.2.3-osc-abcdefg`, which is different to the version this (upstream) repo has.

It would be helpful to be able to override the automatic version generation.

PlagueCZ:
If set during `meson build`, environment variable `DPSERVICE_VERSION` has the ability to override the version generation from current git tag.

Without this variable set, everything should behave as it previously did.

Fixes #773


byteocean (approved):
LGTM. thanks.

PR #777, opened 2026-04-30
Feature remove no conntrack

### LINKED ISSUE #416: Remove --no-conntrack option
After discussion with @guvenc this is maybe something to be done in the future as dpservice does not really work well without it.

There is still a question about removing all `if (!cntrack)`-type checks in the code to eliminate unnecessary conditions. But that may not be possible dues to NULL representing a full conntrack table.

PlagueCZ:
This PR simply removes the option as it has no use anymore.

Fixes #416

### DIFF
--- a/src/nodes/conntrack_node.c
+++ b/src/nodes/conntrack_node.c
@@ -20,6 +20,7 @@
 
 #define NEXT_NODES(NEXT) \
 	NEXT(CONNTRACK_NEXT_LB, "lb") \
+	NEXT(CONNTRACK_NEXT_DHCP, "dhcp") \
 	NEXT(CONNTRACK_NEXT_DHCPV6, "dhcpv6") \
 	NEXT(CONNTRACK_NEXT_DNAT, "dnat") \
 	NEXT(CONNTRACK_NEXT_FIREWALL, "firewall")
@@ -42,7 +43,7 @@ static __rte_always_inline rte_edge_t get_next_index(__rte_unused struct rte_nod
 		if (DP_FAILED(dp_extract_l4_header(df, ipv4_hdr + 1)))
 			return CONNTRACK_NEXT_DROP;
 		if (df->l4_type == IPPROTO_UDP && df->l4_info.trans_port.dst_port == htons(DP_BOOTP_SRV_PORT))
-			return CONNTRACK_NEXT_DNAT;
+			return CONNTRACK_NEXT_DHCP;
 	} else if (df->l3_type == RTE_ETHER_TYPE_IPV6) {
 		dp_extract_ipv6_header(df, ipv6_hdr);
 		if (DP_FAILED(dp_extract_l4_header(df, ipv6_hdr + 1)))
--- a/src/nodes/conntrack_node.c
+++ b/src/nodes/conntrack_node.c
@@ -5,7 +5,6 @@
 #include <rte_graph.h>
 #include <rte_graph_worker.h>
 #include <rte_mbuf.h>
-#include "dp_conf.h"
 #include "dp_error.h"
 #include "dp_flow.h"
 #include "dp_log.h"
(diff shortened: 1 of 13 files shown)


src/nodes/dnat_node.c
```diff
@@ -31,9 +31,6 @@ static __rte_always_inline rte_edge_t get_next_index(__rte_unused struct rte_nod
 	struct dnat_data *dnat_data;
 	union dp_ipv6 nat_ipv6;
 
-	if (!cntrack)
-		goto out;
-
```
PlagueCZ, line 36: This is were BOOTP traffic would jump to `out` because it has no conntrack.

src/nodes/conntrack_node.c
```diff
@@ -42,7 +42,7 @@ static __rte_always_inline rte_edge_t get_next_index(__rte_unused struct rte_nod
 		if (DP_FAILED(dp_extract_l4_header(df, ipv4_hdr + 1)))
 			return CONNTRACK_NEXT_DROP;
 		if (df->l4_type == IPPROTO_UDP && df->l4_info.trans_port.dst_port == htons(DP_BOOTP_SRV_PORT))
-			return CONNTRACK_NEXT_DNAT;
```
PlagueCZ, line 45: This is where BOOTP was jumping to DNAT. As you can see this is before call to `dp_cntrack_handle()`, therefore this flow would be missing conntrack.

src/nodes/dnat_node.c
```diff
@@ -160,7 +157,6 @@ static __rte_always_inline rte_edge_t get_next_index(__rte_unused struct rte_nod
 		return DNAT_NEXT_IPV6_LOOKUP;
 	}
 
-out:
 	if (df->l3_type == RTE_ETHER_TYPE_IPV4)
 		return DNAT_NEXT_IPV4_LOOKUP;
```
PlagueCZ, line 163: And here all BOOTP would go to IPV4_LOOKUP.

src/nodes/ipv4_lookup_node.c
```diff
@@ -31,10 +29,6 @@ static __rte_always_inline rte_edge_t get_next_index(__rte_unused struct rte_nod
 	const struct dp_port *in_port = dp_get_in_port(m);
 	const struct dp_port *out_port;
 
-	// TODO: add broadcast routes when machine is added
-	if (df->l4_type == IPPROTO_UDP && df->l4_info.trans_port.dst_port == htons(DP_BOOTP_SRV_PORT))
-		return IPV4_LOOKUP_NEXT_DHCP;
-
```
PlagueCZ, line 37: All BOOTP will be now redirected to DHCP.

src/nodes/conntrack_node.c
```diff
@@ -42,7 +42,7 @@ static __rte_always_inline rte_edge_t get_next_index(__rte_unused struct rte_nod
...
+			return CONNTRACK_NEXT_DHCP;
```
PlagueCZ, line 45: Contrast with the original code - simply jump directly to DHCP.

PlagueCZ (commented):
As I was first surprised by this change, I tried to explain how I understand the reason behing it.

src/nodes/packet_relay_node.c
```diff
@@ -75,9 +75,6 @@ static __rte_always_inline rte_edge_t get_next_index(__rte_unused struct rte_nod
 	struct dp_flow *df = dp_get_flow_ptr(m);
 	struct flow_value *cntrack = df->conntrack;
 
-	if (!cntrack)
```
byteocean, line 78: the reason why it is ok not to check if `cntrack` is null will be traced back to the 'routing' logic inside conntrack_node.c, which is a few nodes away from the current one.

maybe still worthy of using a simple assertion like `assert(cntrack)` here? (same for the other two places)

PlagueCZ: That is actually a good idea, I put it into all five places I found. Please check.

byteocean: thanks guys. looks good.

byteocean (approved):

PR #778, opened 2026-05-11
Feature/rename dhcp ipv6

### LINKED ISSUE #704: `own_ipv6` in `dp_port` structure
field `own_ipv6` is basically unused:
```
plague@smrcek ~/git/dpservice $ grep -rn own_ipv6 src/
src/nodes/ipv6_nd_node.c:80:            dp_copy_ipv6(&port->iface.cfg.own_ipv6, dp_get_dst_ipv6(req_ipv6_hdr));
plague@smrcek ~/git/dpservice $ grep -rn own_ipv6 include/
include/dp_port.h:31:   union dp_ipv6                   own_ipv6;
```

In contrast with `own_ip` which is used in many places (gRPC, ARP, DHCP, ...).

But there is also `dhcp_ipv6` which is actually used almost like `own_ip` above.

Am I missing something or should these be merged?

PlagueCZ:
This PR simply removes the unused `own_ipv6` in `ipv6_nd_node.c` (it was set and never used). Then it reuses the name for the original `dhcp_ipv6`. This way the name is consistent with the `uint32_t own_ip`

Fixes #704

### DIFF
--- a/include/dp_port.h
+++ b/include/dp_port.h
@@ -29,7 +29,6 @@ struct dp_iface_cfg {
 	uint32_t				neigh_ip;
 	uint8_t					ip_depth;
 	union dp_ipv6			dhcp_ipv6;
-	union dp_ipv6			own_ipv6;
 	uint8_t					ip6_depth;
 	struct dp_ip_address	pxe_ip;
 	char					pxe_str[DP_IFACE_PXE_MAX_LEN];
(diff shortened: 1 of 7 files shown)


peanball (approved):
Confirmed that the field `own_ipv6` was never read and could not be read from outside. Renaming it looks safe.

PR #779, opened 2026-06-10
dp_rte_flow_helpers: Explicit array initialization

mkalcok:
GCC 15 introduces 'unterminated-string-initialization'[0] warning and since we treat warning as errors in our builds, we might as well get ahead of the issue and change from string literals to explicit array initialization.

[0] https://gcc.gnu.org/cgit/gcc/commit/?id=44c9403ed1833ae71a59e84f9e37af3182be0df5

# Proposed Changes

- Use explicit byte array initialization instead of string literals.


mkalcok:
PR from fork seems to cause problems in CI.
Closed in favor of #780

Issue #784, opened 2026-06-21
Crashes on route creation


PlagueCZ:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fa49ab393ea in rte_stack_pop (s=0x0, obj_table=0x7fa4985dc9e8, n=1) at ../lib/stack/rte_stack.h:140

warning: 140    ../lib/stack/rte_stack.h: No such file or directory
[Current thread is 1 (LWP 9)]
#0  0x00007fa49ab393ea in rte_stack_pop (s=0x0, obj_table=0x7fa4985dc9e8, n=1) at ../lib/stack/rte_stack.h:140
#1  stack_dequeue (mp=0x11c669b5c0, obj_table=0x7fa4985dc9e8, n=1) at ../drivers/mempool/stack/rte_mempool_stack.c:59
#2  0x00007fa49d316650 in rte_mempool_ops_dequeue_bulk (mp=0x11c669b5c0, obj_table=0x7fa4985dc9e8, n=1) at ../lib/mempool/rte_mempool.h:793
#3  rte_mempool_do_generic_get (mp=<optimized out>, obj_table=0x7fa4985dc9e8, n=<optimized out>, cache=<optimized out>) at ../lib/mempool/rte_mempool.h:1608
#4  rte_mempool_generic_get (mp=<optimized out>, obj_table=<optimized out>, n=<optimized out>, cache=<optimized out>) at ../lib/mempool/rte_mempool.h:1661
#5  rte_mempool_get_bulk (mp=0x11c669b5c0, obj_table=0x7fa4985dc9e8, n=1) at ../lib/mempool/rte_mempool.h:1696
#6  rte_mempool_get (mp=0x11c669b5c0, obj_p=0x7fa4985dc9e8) at ../lib/mempool/rte_mempool.h:1722
#7  node_alloc (rib=rib@entry=0x11c7894f40) at ../lib/rib/rte_rib.c:91
#8  0x00007fa49d316ac5 in rte_rib_insert (rib=rib@entry=0x11c7894f40, ip=ip@entry=184156216, depth=depth@entry=32 ' ') at ../lib/rib/rte_rib.c:274
#9  0x000055b64732aa3c in dp_add_route (port=0x55b65a638490, vni=<optimized out>, t_vni=207, ip=184156216, t_ip6=t_ip6@entry=0x7fa4985dcae0, depth=<optimized out>) at ../src/dp_lpm.c:101
[...]

PR #786, opened 2026-06-21
Make dp_free_rib() safer

PlagueCZ:
In `dp_free_vni_data()` the call to `dp_create_rib()` can fail. If that happens, the pointer `vni_data->ipv4[socket_id]` will be freed, but still present. Subsequent use of this pointer will result in use-after-free bug.

Since there already was a wrapper for freeing this pointer, I simply added a set to NULL.

Sure in most cases the owning structure is immediately freed, so this could be optimized, but since such code seems to only be called rarely (VNI manipulation), I chose not to optimize, to make it cleaner.

Connected to #784

### DIFF
--- a/src/dp_vni.c
+++ b/src/dp_vni.c
@@ -44,12 +44,18 @@ bool dp_is_vni_route_table_available(uint32_t vni, int type)
 
 static __rte_always_inline void dp_free_rib6(struct dp_vni_data *vni_data)
 {
-	rte_rib6_free(vni_data->ipv6[DP_SOCKETID(vni_data->socket_id)]);
+	int socket_id = DP_SOCKETID(vni_data->socket_id);
+
+	rte_rib6_free(vni_data->ipv6[socket_id]);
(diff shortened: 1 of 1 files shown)


src/dp_vni.c
```diff
@@ -198,13 +204,13 @@ static int dp_reset_vni_data(uint32_t vni, struct dp_vni_data *vni_data)
 	int socket_id = DP_SOCKETID(vni_data->socket_id);
 
 	if (vni_data->ipv4[socket_id]) {
-		rte_rib_free(vni_data->ipv4[socket_id]);
+		dp_free_rib(vni_data);
 		if (DP_FAILED(dp_create_rib(vni, socket_id, vni_data)))
```
byteocean, line 208: if `rte_rib_create` fails, is it better to set both `vni_data->vni` and `vni_data->socket_id` to some invalid values? or simply set all fields of `vni_data` in the failure section inside `dp_create_rib`?

byteocean (approved):

PR #788, opened 2026-06-21
Fix possible double-free of flow table entry

PlagueCZ:
If flow table insertion fails, the rollback code has a bug. Due to the nature of `dp_delete_flow()` which calls `dp_ref_dec()`, the flow value would already be freed by the reference counter hitting zero.

But `dp_delete_flow()` can actually fail, so the result needs to be passed to the caller to decide if `rte_val` still remains to be freed or not.

Connected to issue #784

### DIFF
--- a/include/dp_flow.h
+++ b/include/dp_flow.h
@@ -145,7 +145,7 @@ bool dp_are_flows_identical(const struct flow_key *key1, const struct flow_key *
 
 int dp_get_flow(const struct flow_key *key, struct flow_value **p_flow_val);
 int dp_add_flow(const struct flow_key *key, struct flow_value *flow_val);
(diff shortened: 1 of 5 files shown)


byteocean (commented):
`dp_delete_flow` is also used in [snat_node.c](https://github.com/ironcore-dev/dpservice/blob/b8c1415beb807384e63e55faa0619961a0bb7bc0/src/nodes/snat_node.c#L65) and [dnat_node.c](https://github.com/ironcore-dev/dpservice/blob/b8c1415beb807384e63e55faa0619961a0bb7bc0/src/nodes/dnat_node.c#L83). If we follow the same logic and new function signature, we also need to do early error returning if it fails there.

byteocean (approved):
Thanks for the extra comments, which make things clearer and more precise.

Issue #794, opened 2026-06-29
DHCPv6 Solicit silently dropped when interface has no IPv6 address


byteocean:
When a dpservice interface is created without an IPv6 address`--ipv6 omitted, defaults to ::`, the dpservice DHCPv6 node silently drops incoming DHCPv6 Solicit messages. The Router Advertisement still sets `M=1
  (managed)`, so the guest VM dutifully sends a DHCPv6 Solicit — and then retransmits forever waiting for a reply that never comes. This causes long boot delays and confusing failures.

PR #795, opened 2026-07-06
send dhcp6 response to VM without IPv6 address configured

byteocean:
address #794 

Instead of dropping the Solicit, reply with NoAddrsAvail (RFC 8415 §18.3.2). The guest gets an explicit "no address for you" and stops retrying, while still keeping its link-local address and IPv6 default route from the RA.

There are other approaches such as:
1) Per-interface M bit based on own_ipv6. Requires providing the port into dp_ipv6_fill_ra() (both solicited and periodic paths), and testing that no VF sees the wrong flag. Larger blast radius for the same outcome.

2) Drop RS entirely for no-IPv6 interfaces. Guest waits ~12 s for RA before giving up, then has no IPv6 default route at all.  Also asymmetric with IPv4 (which always ARPs).

To completely disable IPv6 for a VM without IPv6 configured, it is the best to address it via ignition to disable interfaces' IPv6 inside VM.

### DIFF
--- a/src/nodes/dhcpv6_node.c
+++ b/src/nodes/dhcpv6_node.c
@@ -269,6 +269,21 @@ static int generate_reply_options(struct rte_mbuf *m, uint8_t *options, int opti
 	return reply_options_len;
 }
 
+static __rte_always_inline int generate_noaddrs_reply(struct rte_mbuf *m, uint8_t *options, int options_len)
+{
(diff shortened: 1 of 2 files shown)


PlagueCZ (approved):
Code looks good, tests also look good.

Cannot test in practice as OSC does not yet support IPv6, but it also means, that not even breaking changes should affect OSC production.

PR #796, opened 2026-07-06
ci: fix fork PR support in workflow checkouts

byteocean:
Current CI/CD does not support running on forked repo, which is inconvenient for reviewing contributions from outside the org.

 - checkpatch, publish-docker-tester: pass head.sha + head.repo.full_name to actions/checkout so fork branches resolve correctly. Fixes 'branch or tag ... could not be found' failures on fork-triggered PRs.

  - size-label: switch to pull_request_target so the token has write scope on the base repo's labels (fixes 403 Resource not accessible).

 Fork PRs from non-members require maintainer approval (already configured in project setting) before these workflows run, keeping the self-hosted dpdk runner safe.

### DIFF
--- a/.github/workflows/checkpatch.yaml
+++ b/.github/workflows/checkpatch.yaml
@@ -17,7 +17,8 @@ jobs:
     steps:
       - uses: actions/checkout@v6
         with:
-          ref: ${{ github.event.pull_request.head.ref }}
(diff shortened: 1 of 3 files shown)


PlagueCZ (approved):
Statically looking good. OSC does not use these pipelines so cannot verify in practice.

PR #798, opened 2026-07-15
Remove size-label Github Actions workflow

gonzolino:
We are removing the size-label action from all repos, as they add little to no value.


peanball (approved):

maxmoehl:
@byteocean we'll have to fix the docker workflow as we can no longer push from a fork to the registry.

byteocean:
@maxmoehl the changes on the metalnet side will be replicated here, once they get settled there.

PR #801, opened 2026-07-16
Introduce 'ok-to-test' label to guard and run job in base repo's context

byteocean:
fixing failed testing workflow for forked repo:
 
 **Problem**: 
     publish-docker-tester.yml unconditionally pushes to GHCR on every PR. When a PR comes from a fork, GitHub runs the workflow in the fork's restricted context — the GITHUB_TOKEN has no write access to ironcore-dev packages, so the push fails.
     The tester image must be built from PR code and pushed to GHCR before the test job can pull it by SHA. Unlike metalnet, there's no way to skip the push and still run tests.
    
 **Fix**:
    Gate the job on a `ok-to-test` label for fork PRs. When a maintainer applies the label, GitHub runs the workflow as a labeled event — always in the base repo's context with full GHCR write access. Internal PRs (same repo) continue to run automatically.

### DIFF
--- a/.github/workflows/publish-docker-tester.yml
+++ b/.github/workflows/publish-docker-tester.yml
@@ -13,6 +13,7 @@ on:
       - 'docs/**'
       - '**/*.md'
   pull_request:
+    types: [labeled, opened, synchronize, reopened]
     paths-ignore:
       - 'docs/**'
       - '**/*.md'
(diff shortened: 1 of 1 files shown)


mkalcok:
If I understand correctly, the reason for pushing to ghcr registry is to build the image on one machine (gh runner) and then test it on another (self hosted runner). If that's the only reason, we can sidestep the registry upload completely by [uploading the resulting image as a short-lived artifact](https://github.com/actions/upload-artifact) from one job and [download it locally](https://github.com/actions/download-artifact) in the next job.

This way we wouldn't need the access to the secret token that allows upload to registry.

mkalcok:
It's worth noting that [anyone with `triage` permissions can apply labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels#applying-a-label), it's not restricted just to maintainers.

byteocean:
the issue is resolved by #802

PR #802, opened 2026-07-16
Use artifacts to store compiled tester image

byteocean:
Follow the suggestion from @mkalcok, use artifacts to store resulted tester image.

### DIFF
--- a/.github/workflows/publish-docker-tester.yml
+++ b/.github/workflows/publish-docker-tester.yml
@@ -37,26 +37,6 @@ jobs:
           else
            echo "DOCKER_IMAGE_SHA=${{ github.sha }}" >> $GITHUB_ENV
           fi
-      - uses: docker/metadata-action@dc802804100637a589fabce1cb79ff13a1411302 # v6.2.0
-        id: meta
-        with:
-          images: |
-            ghcr.io/${{ github.repository_owner }}/dpservice-tester
-          tags: |
-            type=semver,pattern={{version}}
-            type=schedule
-            type=raw,${{ env.DOCKER_IMAGE_SHA }}
-            type=ref,event=branch
-            type=ref,event=tag
-            type=ref,event=pr
-            type=sha
-          flavor: |
-            latest=${{ github.ref == 'refs/heads/main' }}
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@96fe6ef7f33517b61c61be40b68a1882f3264fb8 # v4.2.0
-        with:
-          image: tonistiigi/binfmt:latest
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-docker-tester.yml
```diff
@@ -71,24 +51,40 @@ jobs:
...
         uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
         with:
           context: .
           platforms: ${{env.platforms}}
-          push: true
-          tags: ${{ steps.meta.outputs.tags }}
-          labels: ${{ steps.meta.outputs.labels }}
+          push: false
+          load: true
+          tags: dpservice-tester:${{ env.DOCKER_IMAGE_SHA }}
```
mkalcok, line 62: I don't think we need the `secrets:` key below, it would likely be empty in a fork anyway, right?

.github/workflows/publish-docker-tester.yml
```diff
@@ -71,24 +51,40 @@ jobs:
...
           target: tester
           secrets: |
             "github_token=${{ secrets.BOT_PAT }}"
+      - name: Save tester image as tar
+        run: docker save dpservice-tester:${{ env.DOCKER_IMAGE_SHA }} | gzip > /tmp/dpservice-tester.tar.gz
+      - name: Upload tester image artifact
```
mkalcok, line 66: Out of curiosity, what is the purpose of pushing the "tester" image to GHCR?

mkalcok (changes requested):
Overall LGTM, thank you. I think there's one left-over secret reference that's no longer needed (see below).

.github/workflows/publish-docker-tester.yml
byteocean: metalnet uses it in its test. [link](https://github.com/ironcore-dev/metalnet/blob/d687c067d379cd28064c17e2f7d1bcfdd88626c1/.github/workflows/test.yml#L34)

mkalcok: ack, thanks.

byteocean: and it should only push image on the main branch

.github/workflows/publish-docker-tester.yml
byteocean: I already removed it, and let's see if it goes through all workflows.

mkalcok (approved):
LGTM, thanks for the updates.

maxmoehl (approved):

PR #810, opened 2026-08-26
Clarify IPv6 host prefix uniqueness in deployment README

markus-hentsch:
The current form of the documentation uses the term "host prefix" without clearly specifying that it is meant to be strictly unique per host, leaving only the term itself suggesting that detail. Since this is in contrast to, e.g., ULA IPv6 address behavior and classic subnet scoping, this should be stated explicitly in the documentation to avoid misconfiguration.

## Summary by CodeRabbit

* **Documentation**
  * Added deployment guidance requiring each host’s loopback IPv6 `/64` prefix to be unique and not shared with other hosts.


mkalcok (approved):
More clarity in the documentation is always welcome. Thank you.

Note though, that the commit needs to be signed off.

markus-hentsch:
> Note though, that the commit needs to be signed off.

Done.

mkalcok:
Thanks.

PR #816, opened 2026-09-16
Enforce firewall rules with per-interface enablement

guvenc:
## Summary
Firewall rules were evaluated but a drop decision was ignored, so no rule was ever enforced. This PR enforces them and makes the firewall controllable per interface.

## Changes
- **Enforcement:** Rules are enforced in both directions. A direction with no rules allows all traffic; once a direction has a rule, non-matching traffic is dropped.
- **Rule matching:** IP prefixes are now checked for every protocol, including ICMP and match-any rules. Port ranges apply only to TCP/UDP, and type/code only to ICMP.
- **Per-interface state:** The firewall is enabled by default on each interface. It can be read and changed with the new `GetFirewallParams` / `SetFirewallParams` gRPC calls, and is also returned by `GetInterface` / `ListInterfaces`.
- **Clients:** The new calls are exposed in the Go client, in `dpservice-cli` (`get fwparams` / `set fwparams`) and in the pytest gRPC client.
- **Rule hit counts:** A new telemetry command, `/dp_service/firewall/rule_hits`, returns per-rule hit counters. The Prometheus exporter publishes them as `dps_firewall_rule_hits_total`.
- **Conntrack fix for NAT:** Only translated (south-north) flows are marked as NAT in conntrack keys. Before this, replies between a NAT'd and a non-NAT'd VM missed their conntrack entry and were dropped by the ingress rules.
- **HA:** Flows taken over through NAT sync now get a firewall decision.
- **Monitoring fix:** The monitoring event queue is now multi-producer, because events are sent from several threads.

## Behavior change
- Existing firewall rules now take effect: an interface with any rule for a direction drops traffic in that direction that matches none of them.

## Testing
- New conntrack tests for NAT'd interfaces:
  - VM to VM on the same host, VM to a VM on another host, and VM to the internet
  - TCP, ICMP, IPv6 and NAT64
  - HA failover
- Existing tests adjusted to the enforced firewall.
- All local suites pass (default, port redundancy, HA, flow timeout).

## Summary by CodeRabbit

* **New Features**
  * Added CLI and API support to view and update firewall state per interface.
  * Added firewall rule-hit telemetry through `dps_firewall_rule_hits_total`.
  * Interface details now report the current firewall state.

* **Bug Fixes**
  * Improved protocol-specific rule matching, disabled-firewall handling, and synchronized NAT flow filtering.
  * Restored firewall drops for blocked traffic and improved rule-hit accuracy.
  * Expanded coverage for firewall state, ICMP matching, telemetry, and NAT scenarios.


docs/deployment/help_dpservice-bin.md
```diff
@@ -23,6 +23,7 @@
 | --log-format | FORMAT | set the format of individual log lines (on standard output) | 'text' (default) or 'json' |
 | --grpc-port | PORT | listen for gRPC clients on this port |  |
 | --flow-timeout | SECONDS | inactive flow timeout (except TCP established flows) |  |
+| --fwall-tel-interval | SECONDS | minimal interval between firewall rule hits telemetry refreshes |  |
```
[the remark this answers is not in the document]
PlagueCZ: This is true for --flow-timeout also and one other setting already. Needs a separate PR that changes the generator.

src/dp_firewall.c
```diff
@@ -80,14 +80,14 @@ int dp_list_firewall_rules(const struct dp_port *port, struct dp_grpc_responder
...
+	uint32_t dst_port_lower, dst_port_upper;
 	union dp_ipv6 r_src_ip6;
 	union dp_ipv6 r_dest_ip6;
-	uint8_t protocol = df->l4_type;
-	uint16_t dest_port = 0;
-	uint16_t src_port = 0;
+	uint16_t dest_port;
+	uint16_t src_port;
 
+	/* Protocol gate: the rule must be a wildcard or match the packet's L4 protocol. */
```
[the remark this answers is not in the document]
PlagueCZ: Not sure what the issue here is. Unknown protocol returns false, i.e. no rule match, i.e. dropped. Since ther eis no way to even allow such protocol, this is the desired behavior and the firewall is allow-based.

PlagueCZ (changes requested):
I did a partial review, I am still unsure with the telemetry and tests, but I want to put the firewall code review here.

The firewall itself has been tested in real environment a little bit. That is how we found the conntrack problem, which is also fixed here.

My current issue is the telemetry is so different than all previous ones, that I need to look further.

src/dp_firewall.c
```diff
@@ -185,25 +198,23 @@ enum dp_fwall_action dp_get_firewall_action(struct dp_flow *df,
...
-	if (egress_action != DP_FWALL_ACCEPT)
+	enum dp_fwall_action egress_action = DP_FWALL_ACCEPT;
+	enum dp_fwall_action ingress_action = DP_FWALL_ACCEPT;
+
+	/* A PF holds no firewall rules, so each direction is only evaluated on its VF side.
+	 * Both directions are evaluated even when one of them already decided to drop, so that
+	 * the per-rule hit counters account for every matching rule.
+	 */
+	if (!in_port->is_pf)
+		egress_action = dp_get_directional_action(df, &in_port->iface.fwall_head, DP_FWALL_EGRESS);
```
byteocean, line 209: Inside `dp_get_directional_action`, `dp_is_matched_in_fwall_list` is called and `rule->stats.rule_hit++` is executed if a rule matching happens. This sequence of calling seems to be invoked regardless if the below `.fwall_state` equals to `DP_FWALL_ENABLED` or not. That is said, regardless if a firewall is enabled or not, `stats.rule_hit` keeps increasing. Is it intended behaviour or `stats.rule_hit` should stop increasing when fwall is disabled?

src/dp_cntrack.c
```diff
@@ -207,6 +208,49 @@ static __rte_always_inline struct flow_value *flow_table_insert_entry(struct flo
...
+		dp_ipv6_from_ipaddr(&df->dst.dst_addr6, &key->l3_dst);
+	} else {
+		df->src.src_addr = htonl(key->l3_src.ipv4);
+		df->dst.dst_addr = htonl(key->l3_dst.ipv4);
+	}
+	if (key->proto == IPPROTO_TCP || key->proto == IPPROTO_UDP) {
+		df->l4_info.trans_port.src_port = htons(key->src.port_src);
+		df->l4_info.trans_port.dst_port = htons(key->port_dst);
+	} else if (key->proto == IPPROTO_ICMP || key->proto == IPPROTO_ICMPV6) {
+		df->l4_info.icmp_field.icmp_type = (uint8_t)key->src.type_src;
```
byteocean, line 230: `flow_key` does not have the `icmp_code` field, which is used inside `dp_is_rule_matching`. Allowing icmp packets with the same type (e.g., error type) and a specific code seems to allow following icmp packets with the same type with different codes because flow_key remains the same, and it may not be the intent of a fwall rule installation call specifying a icmp code. To resolve this potential inconsistency, either `flow_key` needs to include `icmp_code`, or `icmp_code` needs to be always `DP_FWALL_MATCH_ANY_ICMP_CODE`, or remove this field from this gRPC protocol.

byteocean: Adding `icmp_code` to `flow_key` can be tricky because unknown type / code of the replying packet for a icmp request. If we are fine with larger granularity in terms of `icmp_code` in firewall, it is just easy to remove it from the proto file.

src/dp_firewall.c
PlagueCZ: That is intenional. The reason is::
 - you have some rules
 - nothing works
 - you disable the firewall but keep rules
 - packets now work, but you can also see what *would* happen if you enabled it

guvenc:
The telemetry part of this PR will get a re-work.

